Issue #9, opened 2018-12-13
UserInfo is set as Principle but does not satisfy Spring "Spec"


nenaraab:
### Problem 
`UserInfo` is set as Principle by  the `XsuaaUserAuthenticationInfo`, which implements `org.springframework.security.core.Authentication`.

The Spring framework often calls `SecurityContextHolder.getContext().getAuthentication().getPrincipal()`  and expects either a Principal of type `org.springframework.security.core.userdetails.UserDetails` or expects (more implicitly) that the `toString()` method is implemented in such way, that it returns a (unique) userName...

The simplest fix would be to let the `UserInfo` in the commons lib override the `toString()` method:
```java
/** 
* As principle it should return a unique user name as required by Spring Security.
@Override
public String toString() {
   return getLogonName();
}
```
> Note: This was also an incompatible change to SAPUserDetails, which was provided with the sap internal library security-commons version `0.31.0`.

nenaraab:
@mwdb 
Unfortunately your pull request does not fix the issue. It must be sth, that can be used as database entry for the Spring ACL tables, e.g. `getLogonName()`.

nenaraab:
see 
https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/Token.java 
implements now `org.springframework.security.core.userdetails.UserDetails`

PR #10, opened 2018-12-13
UserInfo is set as Principle but does not satisfy Spring Spec

nenaraab:
@mwdb 
Question: is it possible, that the JWT claim "origin" is not filled? Because in the `UserInfoTestEndUser` test the jwt token had no "origin" claim.

This PR fixes the issue https://github.com/SAP/cloud-security-xsuaa-integration/issues/9

### DIFF
--- a/spring-xsuaa/src/main/java/com/sap/xs2/security/container/UserInfo.java
+++ b/spring-xsuaa/src/main/java/com/sap/xs2/security/container/UserInfo.java
@@ -1,34 +1,32 @@
 package com.sap.xs2.security.container;
 
-import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.Base64;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
+import com.sap.xsa.security.container.XSTokenRequest;
+import com.sap.xsa.security.container.XSUserInfo;
+import net.minidev.json.JSONArray;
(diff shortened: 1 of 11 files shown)


spring-xsuaa/src/main/java/com/sap/xs2/security/container/UserInfo.java
```diff
@@ -630,6 +631,29 @@ private URI replaceSubdomain(URI uri, String subdomain) {
...
+	@Override
+	public String getName() {
+		String principalName = null;
+		try {
+			if (GRANTTYPE_CLIENTCREDENTIAL.equals(getGrantType())) {
+				principalName = getClientId();
+			} else {
+				principalName = getLogonName();
+				if (getOrigin() != null) {
+					principalName = getOrigin() + "." + principalName;
```
mwdb: The "." is a very common seperator. I'd rather use something like /,# or |

nenaraab: @mwdb 
Thanks, Martijn! I've corrected the delimiter to "/"... maybe in future we can make it configurable...

nenaraab:
@mwdb 
as discussed I've updated my pull request.
Now, UserInfo does not implement XSPrincipal but UserDetails from spring security.
Furthermore i've refactored UserInfoTestUtil and I've fixed another issue, which was documented here #11 

Can you please review?

PR #22, opened 2019-01-02
Make XsuaaAudienceValidator more robust

nenaraab:
added further test cases to eliminate NullPointer exception and other issues, when the claim "scope" is null or empty as part of the JWT token.

fixes #22


mwdb:
Hi Nena,

Thanks for the pull request.
Two comments:
1) Can you update the changelog (https://github.com/SAP/cloud-security-xsuaa-integration/blob/develop/CHANGELOG.md)
2) The scope parameter is mandatory for uaa. UAA will not issue an access token with an empty scope or no scopes.
Anyway as it improves it I would merge it after you added the changes in the changelog.

Regards,

Martijn

nenaraab:
Okay, will close this one and open a PR on "master" branch instead

PR #27, opened 2019-01-03
Minor improvements with version 1.1.1

nenaraab:
## Version 1.1.1
* Eases enhancement of TokenAuthenticationConverter (https://github.com/SAP/cloud-security-xsuaa-integration/issues/23)
* Makes XsuaaAudienceValidator more robust (https://github.com/SAP/cloud-security-xsuaa-integration/issues/21)
* XSTokenRequest accepts custom RestTemplate (https://github.com/SAP/cloud-security-xsuaa-integration/issues/25)

PR #28, opened 2019-01-03
Develop branch: version 1.2.0

nenaraab:
@see PR 
https://github.com/SAP/cloud-security-xsuaa-integration/pull/27

This PR contains the same commits to keep develop and master branch in synch


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/TokenAuthenticationConverter.java
```diff
@@ -1,42 +1,58 @@
...
+	public final AbstractAuthenticationToken convert(Jwt jwt) {
+		return new AuthenticationToken(appId, jwt, extractAuthorities(jwt));
+	}
 
     protected Collection<GrantedAuthority> extractAuthorities(Jwt jwt) {
-        Collection<String> scopes = this.getScopes(jwt);
-        return scopes.stream().map(authority -> authority).map(SimpleGrantedAuthority::new).collect(Collectors.toList());
-    }
+        Collection<String> scopeAuthorities = getScopes(jwt);
+        Collection<String> otherAuthorities = extractAuthorities(new TokenImpl(jwt, appId));
```
mwdb: Libraries only validate the scope field.
All user tokens (password grant, SAML, authorization grant) only contain scopes, the client credential tokens contain scopes and authorities. Scopes are filtered by using /oauth/token?scope=... but authorities are not.
Therefore validation must only look into the scope field.

nenaraab:
@mwdb 
- Which libraries do you mean?
- The default, as we deliver is with scopes only - no change here
- With this change we only offer applications a safe way to enhance the list by other (custom) authorities such as xs user attributes. Have a look into [this example](https://github.wdf.sap.corp/CPSecurity/cp-application-security/blob/master/spring-security-acl/src/main/java/com/sap/cp/appsec/security/CustomTokenAuthorizationsExtractor.java)
- Scopes and other authorities are clearly separated as documented in this [test case](https://github.com/SAP/cloud-security-xsuaa-integration/pull/28/commits/02616ddc11a10d084b7a7b2f5bcb5ccd0d67b0c1#diff-55b585569b10c1eeaacc5e8f11b9598aR62)

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/TokenAuthenticationConverter.java
nenaraab: No, the extractAuthorities method returns an empty collection:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/d6e71ea7944d1e08fc538158497203c1b83c7fa9/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/TokenAuthenticationConverter.java#L51

PR #56, opened 2019-02-13
add formatter for spring-xsuaa-it

mwdb:


mwdb:
Hi Nena,

your PR was missing the formatter goal for the spring-xsuaa-it subproject.
spring-xsuaa-it does not inherit from the parent to ensure it never ends up in maven central by accident.
Without this change, man formatter:format fails on spring-xsuaa-it.

Regards,

Martijn

PR #70, opened 2019-04-19
extractToken should support oauth and/or client credentials

nenaraab:
With this change two kind of tokens are supported: Oauth and basic/client-credentials.

E.g. in case the "Authorization" header is not provided, and basic or client-credentials is supported it tries to return the password or client-credentials token.


nenaraab:
@mwdb 
can you please do a review?

mwdb:
basic and JWT always use the authorization header.
Either:
- Authorization: basic base64(username:password)
- Authorization: Bearer <token>

Also the change is in case a null/emty string value is returned, the checks continue

Issue #62, opened 2019-03-27
Xsuaa Audience Validator


nenaraab:
[allowedAudiences()](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidator.java#L54)
- should not return blanks ("") as audience in the list (this can happen, if the xsuaa.xsappname is "")
- should not return duplicates

PR #71, opened 2019-04-26
Fix allowed audiences

nenaraab:
[XsuaaAudienceValidator.allowedAudiences()](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidator.java#L54)
- should not return blanks ("") as audience in the list (this can happen, if the xsuaa.xsappname is "")
- should not return duplicates

Related issue:
#62


nenaraab:
@mwdb 
can you please perform a review?
Thanks :-)

PR #90, opened 2019-06-13
Error fixing

alstolten:
I fixed the links pointing to the raw files to make it possible to download them.

### DIFF
--- a/troubleshooting/logcollector/README.md
+++ b/troubleshooting/logcollector/README.md
@@ -12,7 +12,7 @@ Then you can replay your scenario. After that the script restores the log levels
 - Powershell or Bash
 
 ## How to use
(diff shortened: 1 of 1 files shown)


alstolten:
Hi Nena,

they were relative links before (and pointing nowhere btw, my fault). However the problem is that the file behind a relative link is html formatted and not a executable script (and therefore not "downloadable" by right-clicking and saving). With pointing to the raw file I would like to avoid this problem.

Is there another way? (releases might not be viable in terms of only one or two script files)

Best regards,
Alessandro

nenaraab:
okay, that's an argument. I'm not aware on how to refer to the "raw" document while using a relative link.

mwdb:
@nenaraab : Once you approve, I'll merge the change

PR #94, opened 2019-06-24
Update pom.xml version to 1.5

nenaraab:
- adds update script
- updates versions everywhere (from 1.4.0 -> to 1.5.0)
- as of now the samples must be updated manually


nenaraab:
already integrated in master

PR #97, opened 2019-06-28
Use jku of token to retrieve keys

firnkes:
Use jku of token to retrieve token keys instead of building url.

Aligns this library with other client libs.


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java
```diff
@@ -47,32 +53,72 @@
...
 		} catch (ParseException ex) {
 			throw new JwtException("Error initializing JWT decoder:" + ex.getMessage());
 		}
+
+		String jku = (String) jwt.getHeader().toJSONObject().getOrDefault("jku", null);
+		String kid = (String) jwt.getHeader().toJSONObject().getOrDefault("kid", null);
+		String uaadomain = xsuaaServiceConfiguration.getUaaDomain();
+
+		try {
+			canDoOnlineVerification(jku, kid, uaadomain);
```
mwdb: Can you change the name? I understand onlineVerification as calling the /check_token endpoint. This is not meant, but rather getting the keys from /token_keys

firnkes: done, renamed to VerifyWithOnlineKey

firnkes:
Confilicts are solved, code is formatted.

PR #98, opened 2019-07-01
Use jku of token to retrieve keys

firnkes:
Use jku of token to retrieve token keys instead of building url.

Aligns this library with other client libs.
#97


nenaraab:
@firnkes
As discussed, test lib needs to remain compatible.

PR #74, opened 2019-05-02
Initialization of Spring SecurityContext for non HTTP requests

nenaraab:
In case of non-HTTP requests, you may need to initialize the Spring SecurityContext with a JWT token you've received from a message or you've requested from XSUAA directly.

**Prerequisite**:   
You need to initialize a `JwtDecoder` bean, as explained [here](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java), using the `XsuaaJwtDecoderBuilder` class.

**Then, initialize Security Context**:
```
@Autowired
JwtDecoder jwtDecoder;

public void initializeSecurityContext() {
    Jwt jwtToken = jwtDecoder.decode(<your encoded jwt token String>);
    SecurityContext.init(XSAPPNAME, jwtToken, true);
}
```

By default the SecurityContext is initialized for your current Thread. I.e. you can access the Token from everywhere in your code, that runs in the same Thread:
```
Token token = (Token)SecurityContextHolder.getContext().getAuthentication().getPrincipal();
```

**Finally, clear Security Context**:
Note that `SecurityContextHolder` stores the information in `ThreadLocal`s. That means you should remove current thread's value at the end:
```
SecurityContext.clear();
```

PR #105, opened 2019-07-08
SecurityContext.init method

nenaraab:
With version 1.4.0, Pull Request #74 the [SecurityContext ](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/xs2/security/container/SecurityContext.java) was enhanced by a method to initialize the SecurityContextHolder with a JWT token as documented here:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/README.md#setup-security-context-for-non-http-requests

I'm not that happy with the released API (as of version 1.4.0) because the JWT token must not necessarily be a validated one. Only when people follows the documentation, then it is guaranteed. Please have a look at my code change proposal.

Overload SecurityContext.init method:
`SpringSecurityContext.init(String encodedJwtToken, JwtDecoder jwtDecoder, AuthoritiesExtractor authoritiesExtractor)`

From customer perspective this would be also a simplification.

### incompatible change:
- we face issues, when microservice uses SAP-libs, that use sap-internal container lib, that's why we need to move our functionality to a separate class, namely `SpringSecurityContext`

### DIFF
--- a/spring-xsuaa/src/main/java/com/sap/xs2/security/container/SecurityContext.java
+++ b/spring-xsuaa/src/main/java/com/sap/xs2/security/container/SecurityContext.java
@@ -1,13 +1,15 @@
 package com.sap.xs2.security.container;
 
+import com.sap.cloud.security.xsuaa.extractor.AuthoritiesExtractor;
+import com.sap.cloud.security.xsuaa.extractor.LocalAuthoritiesExtractor;
+import com.sap.cloud.security.xsuaa.token.Token;
 import com.sap.cloud.security.xsuaa.token.TokenAuthenticationConverter;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.util.Assert;
-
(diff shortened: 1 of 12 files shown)


nenaraab:
@strehle 
can you please perform an initial review?

nenaraab:
@mwdb can you please perform a review?

spring-xsuaa/src/main/java/com/sap/xs2/security/container/SecurityContext.java
```diff
@@ -32,36 +34,54 @@ static public UserInfo getUserInfo() throws UserInfoException {
...
+	 * @param extractLocalScopesOnly
+	 *            true when {@link Token#getAuthorities()} should only extract local
+	 *            scopes. Local scopes means that non-application specific scopes
+	 *            are filtered out and scopes are returned without appId prefix,
+	 *            e.g. "Display".
+	 * @deprecated use method
+	 *             {@link SpringSecurityContext#init(String, JwtDecoder, AuthoritiesExtractor)}
+	 *             instead
+	 */
+	@Deprecated
```
mwdb: Why does this need to be deprecated?

nenaraab:
Hi @mwdb 
Because it should not be used any longer :-) ! If people use it as documented, it’s fine. But if they just decode the token w/o validation, finally the security context will contain an invalid jwt...

Issue #112, opened 2019-07-18
Support for Spring projects without Boot


maxvader:
Hello, I have to mantain a project that is configured to produce a War to be run in a Tomcat environment and internally is a classic Spring application with Spring Security but not Boot.
So now I'm stuck in the situation where I use sap java build pack _but_ I use Spring Security and not Boot.
Neither of the 2 solutions you provide is good for me, I tried to use the artifacts for spring security, starting some beans manually, but it doesen't work.
Can you create a classic Spring solution or simply provide some configuration instructions?
I can't switch to Spring Boot ATM.

nenaraab:
Dear @maxvader 

if you are not interested into the test/mock spring-xsuaa libs, then i have good news for you. 
You should be able to use it without Spring-Boot.

Pls, try first with version 1.4.0:
```xml
                <properties>
	        	<sap.cloud.security.version>1.4.0</sap.cloud.security.version>
        	</properties>
                 <dependency>
			<groupId>org.springframework.security</groupId>
			<artifactId>spring-security-oauth2-jose</artifactId>
			<version>5.1.1.RELEASE</version>
		</dependency>
		<dependency>
			<groupId>org.springframework.security</groupId>
			<artifactId>spring-security-config</artifactId>
			<version>5.1.3.RELEASE</version>
		</dependency>
		<dependency>
			<groupId>org.springframework.security</groupId>
			<artifactId>spring-security-oauth2-resource-server</artifactId>
			<version>5.1.1.RELEASE</version>
		</dependency>
		<dependency>
			<groupId>com.sap.cloud.security.xsuaa</groupId>
			<artifactId>spring-xsuaa</artifactId>
			<version>${sap.cloud.security.version}</version>
		</dependency>
               <dependency> <!-- as of version 1.5.0-->
                      <groupId>org.apache.logging.log4j</groupId>
                      <artifactId>log4j-to-slf4j</artifactId>
                      <version>2.11.2</version>
               </dependency>
 ```

and then configure OAuth server in your Security Configuration class:
```java
@Configuration
@EnableWebSecurity
@PropertySource(factory = XsuaaServicePropertySourceFactory.class, value = {""})
public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
    
    @Autowired
    XsuaaServiceConfigurationDefault xsuaaServiceConfiguration;

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        // @formatter:off
        http
[...]

PR #115, opened 2019-07-26
Provide Spring Boot Starter: xsuaa-spring-starter

nenaraab:
With this PR application developers can include spring-xsuaa library with all dependencies with one dependency and leverage auto-configuation as documented [here](https://github.com/SAP/cloud-security-xsuaa-integration/tree/spring-starter/spring-xsuaa#maven-dependencies).

- These samples are adapted accordingly:
  - spring-security-basic-auth
  - spring-security-xsuaa-usage

Furthermore this PR adds Spring-Context tests to the two above sample apps.

Auto-configuration is only enabled, when applications make use of Spring Boot starters. This change was necessary in order to support plain Spring projects (see also [here](https://github.com/SAP/cloud-security-xsuaa-integration/issues/112)). The incompatibility is documented as well.


nenaraab:
@mwdb and another one for review :-)

firnkes:
looks good to me.

nenaraab:
Hi @TheFonz2017,
please my adaptions :-)
Best regards,
Nena

spring-xsuaa/pom.xml
```diff
@@ -42,7 +42,7 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure</artifactId>
 			<version>${spring.boot.version}</version>
-			<scope>provided</scope>
+			<optional>true</optional>
```
mwdb: Hm...this adds a dependency to a Spring version.  This implies every time Spring updates a new version, this lib also needs to update. Why is this needed?

PR #116, opened 2019-07-29
Xsuaa Token flows

nenaraab:
Changes:
- Added XsuaaTokenFlows APIs
- Added tests
- Added sample code in spring-security-xsuaa-usage
- Added migration implementation of old Token.requestToken() API and marked old APIs as deprecated.

Origin PR:
- #110 

Referenced PR
- #124 

TODOS:
- [X] Fix Decoder Issue
- [X] Sample / Documentation


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlow.java
```diff
@@ -0,0 +1,238 @@
...
+		Assert.notNull(tokenDecoder, "TokenDecoder must not be null.");
+
+		this.restTemplate = restTemplate;
+		this.tokenDecoder = tokenDecoder;
+
+		URI tokenEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/token").build().toUri();
+		URI authorizeEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/authorize").build().toUri();
+		URI keySetEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/token_keys").build().toUri();
+
+		this.request = new XsuaaTokenFlowRequest(tokenEndpoint, authorizeEndpoint, keySetEndpoint);
```
nenaraab: call the other constructor

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/NimbusTokenDecoder.java
```diff
@@ -0,0 +1,43 @@
...
+
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see
+	 * com.sap.cloud.address.service.xsuaa.tokenflows.TokenDecoder#setJwksURI(java.
+	 * net.URI)
+	 */
+	@Override
+	public void setJwksURI(URI keySetUri) {
```
nenaraab: For me this is a critical design: 
- XsuaaAutoConfiguration configures a static instance of NimbusTokenDecoder (default BEAN has static scope) .... but what if another parallel Thread (thread-2) overwrites the JwksURI in the meanwhile, while the first thread (thread-1) calls the decode() method?

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/RefreshTokenFlow.java
```diff
@@ -0,0 +1,237 @@
...
+	RefreshTokenFlow(RestTemplate restTemplate, VariableKeySetUriTokenDecoder tokenDecoder, URI xsuaaBaseUri) {
+		Assert.notNull(restTemplate, "RestTemplate must not be null.");
+		Assert.notNull(tokenDecoder, "TokenDecoder must not be null.");
+		Assert.notNull(xsuaaBaseUri, "XSUAA base URI must not be null.");
+
+		URI tokenEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/token").build().toUri();
+		URI authorizeEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/authorize").build().toUri();
+		URI keySetEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/token_keys").build().toUri();
+
+		this.restTemplate = restTemplate;
```
nenaraab: call RefreshTokenFlow constructor to avoid duplicate code

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/UserTokenFlow.java
```diff
@@ -0,0 +1,301 @@
...
+	UserTokenFlow(RestTemplate restTemplate, RefreshTokenFlow refreshTokenFlow, URI xsuaaBaseUri) {
+		Assert.notNull(restTemplate, "RestTemplate must not be null.");
+		Assert.notNull(refreshTokenFlow, "RefreshTokenFlow must not be null.");
+		Assert.notNull(xsuaaBaseUri, "XSUAA base URI must not be null.");
+
+		URI tokenEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/token").build().toUri();
+		URI authorizeEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/oauth/authorize").build().toUri();
+		URI keySetEndpoint = UriComponentsBuilder.fromUri(xsuaaBaseUri).path("/token_keys").build().toUri();
+
+		this.restTemplate = restTemplate;
```
nenaraab: call UserTokenFlow constructor to avoid duplicate code

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlow.java
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	/**
+	 * Executes the token flow and returns a JWT token from XSUAA.
+	 * 
+	 * @return the JWT token generated by XSUAA.
+	 * @throws TokenFlowException
+	 *             in case of token flow errors.
+	 */
+	public Jwt execute() throws TokenFlowException {
```
nenaraab: why do you not return XsuaaToken instead of Jwt - this seems to be inconsistent

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/XsuaaTokenFlowRequest.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.util.Assert;
+
+/**
+ * An internal token exchange request capturing data by the token flow builders
+ * {@link ClientCredentialsTokenFlow} and {@link UserTokenFlow}.
+ */
+class XsuaaTokenFlowRequest {
```
nenaraab: why does it NOT implement com.sap.xsa.security.container.XSTokenRequest?

nenaraab:
@TheFonz2017 
a short summary of the PR review:

- we need to use https://github.com/google/gson instead of jackson (jackson requires too many security patches)
- we should think about using WebClient instead of RestTemplate (can also be used by reactive apps)
- [critical] NimbusTokenDecoder/VariableKeySetUriTokenDecoder as auto-configured in XsuaaAutoConfiguration is not "thread-safe":
  XsuaaAutoConfiguration configures a static instance of NimbusTokenDecoder (default BEAN has static scope) .... but what if another parallel Thread (thread-2) overwrites the JwksURI in the meanwhile, while the first thread (thread-1) calls the decode() method?
  -  we should try to 
- move documentation from samples/spring-security-xsuaa-usage/Readme.md into spring-xsuaa/Readme.md
- for consistency: instead of returning Jwt, why not returning Token? 
- SecurityContextHolder ->

samples/spring-security-xsuaa-usage/README.md
```diff
@@ -60,3 +60,65 @@ After deployment, the AppRouter will trigger authentication automatically when y
...
+XSUAA introduces a new API to support the following token flows:
+
+* Authorization Code Grant
+* User Token Grant
+* Client Credentials Token Grant
+* Refresh Token Grant
+
+Authorization Code Grant is a flow that involves the browser and is therefore triggered by an API gateway (e.g. AppRouter). The other flows, however, may need to be triggered programmatically, e.g. to exchange one token for another or refresh a token, if it is about to expire.
+
+To that end, we now provide the `XsuaaTokenFlows` class, which serves as a factory for the different flows. This deprecates the existing `XsuaaToken.requestToken(XSTokenRequest)` API.
```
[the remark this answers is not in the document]
TheFonz2017: What would you prefer instead? 😉 
And: what's the problem with "we"? 
A little bit of "we" now and then might be a good thing for all of us... 😄

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/TestConstants.java
```diff
@@ -0,0 +1,14 @@
+package com.sap.cloud.security.xsuaa.token.flows;
+
+import java.net.URI;
+
+//@formatter:off
+interface TestConstants {
+    String clientId          = "sb-spring-netflix-demo!t12291";
+    String clientSecret      = "2Tc2Xz7DNy4KiACwvunulmxF32w=";
+    URI xsuaaBaseUri         = URI.create("https://d056076-sub1.authentication.eu10.hana.ondemand.com/");
```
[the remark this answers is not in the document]
TheFonz2017: What is a d-user? 😇 
This is just an arbitrary prefix used to make the URL unique...

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlow.java
TheFonz2017: How do you suggest I do that? The other constructor checks that the parameters are not null. And that is so on purpose. If I did not check the last two parameters for null values, I would not have to override the constructor in the first place. But I think it makes sense to have both constructors and also to check if the parameters are not null.

samples/spring-security-xsuaa-usage/README.md
```diff
@@ -60,3 +60,65 @@ After deployment, the AppRouter will trigger authentication automatically when y
...
+                .client(clientId)
+                .secret(clientSecret)
+                .execute();
+```
+
+In case you have a refresh_token value and want to refresh an existing token with it, you can do the following:
+
+```java
+Jwt refreshToken = xsuaaTokenFlows.refreshTokenFlow(URI.create(baseUrl))
+                    .refreshToken("Your refresh token goes here. You get this from the OAuth server.")
```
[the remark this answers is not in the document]
TheFonz2017: I considered both "dummy token" and an encoded / generated UUID string. 
Both options did not seem like good options for me, since they do not make clear where the token comes from. Also, a refresh token - afaik - has no standardized format. Its just a random value that is used by the server to make the match between a client, a token and a refreshed token. So, to me, this was the best option, also since it will fail at least with an error messag containing this "slightly more verbose" dummy token. 

Not sure why this confused you, though.

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java
```diff
@@ -39,4 +44,45 @@ public XsuaaServiceConfiguration xsuaaServiceConfiguration() {
...
+	/**
+	 * Creates a {@link VariableKeySetUriTokenDecoder} instance based on a
+	 * {@code NimbusJwtDecoderJwkSupport} implementation which is used by the
+	 * {@link XsuaaTokenFlows} bean.
+	 * 
+	 * @return the {@link VariableKeySetUriTokenDecoder} instance.
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	public VariableKeySetUriTokenDecoder xsuaaTokenDecoder() {
```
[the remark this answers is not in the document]
TheFonz2017: The answer is given in class `VariableKeySetUriTokenDecoder`:

```
/**
 * Token decoder interface to provide for 
 * dependency injection of actual token decoder
 * implementation. 
 * 
 * This interface is used to implement a token decoder
 * that can decode tokens from various sources.
 * Since decoding requires the public keys of the token's origin
 * the {@code keySetUri} needs to be specified before decoding.
 * 
 * <b>Note:</b> the standard JwtDecoder exposed as a bean in class 
 * {@link XsuaaResourceServerJwkAutoConfiguration} is referring to 
 * a single Key Set URI, only. This interface defines an API for a
 * JwtDecoder which can be given changing key set URIs.
 */
```
The `XsuaaResourceServerJwkAutoConfiguration.java` uses a keyset URI which (may be tenant-specific but) always points to XSUAA. The Token Flows API can potentially be used with another keyset endpoint / OAuth server (which is why the Token Flows all have two constructors, one of which is taking the actual endpoints to communicate with). Therefore, we need another TokenDecoder here, and the name prefix `VariableKeySetUri...` tried to express that.

nenaraab:
@mwdb 
do you know why this is needed?

https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/XsuaaTokenExchanger.java#L70

```java
if (tokenSubdomain != null && tokenRequestSubdomain != null && !tokenSubdomain.equals(tokenRequestSubdomain)) {
			tokenRequest.setTokenEndpoint(replaceSubdomain(tokenRequest.getTokenEndpoint(), tokenSubdomain));
		}
```

samples/spring-security-xsuaa-usage/README.md
```diff
@@ -60,3 +60,65 @@ After deployment, the AppRouter will trigger authentication automatically when y
...
+                    .client(clientId)
+                    .secret(clientSecret)
+                    .execute();
+
+logger.info("Got the refreshed token: {}", refreshToken.getTokenValue());
+logger.info("You could now inject this into Spring's SecurityContext, using: SecurityContextHolder.getContext().setAuthentication(...).");
```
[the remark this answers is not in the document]
TheFonz2017: Are you sure? Spring initializes a SecurityContext already. Why would one want re-initialize it? The intention here is to reuse the SecurityContext "vehicle" of Spring Security and simply replace the old token with the refreshed one.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/SystemTests.java
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.web.client.RestTemplate;
+
+import com.sap.cloud.security.xsuaa.token.flows.ClientCredentialsTokenFlow;
+import com.sap.cloud.security.xsuaa.token.flows.NimbusTokenDecoder;
+import com.sap.cloud.security.xsuaa.token.flows.TokenFlowException;
+import com.sap.cloud.security.xsuaa.token.flows.VariableKeySetUriTokenDecoder;
+
+public class SystemTests {
```
nenaraab: asumme that TestConstants refer to valid Xsuaa urls

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlow.java
```diff
@@ -0,0 +1,238 @@
...
+		HttpEntity<Void> requestEntity = new HttpEntity<>(headers);
+
+		URI requestUri = builder.build().encode().toUri();
+
+		@SuppressWarnings("rawtypes")
+		ResponseEntity<Map> responseEntity = restTemplate.postForEntity(requestUri, requestEntity, Map.class);
+
+		HttpStatus responseStatusCode = responseEntity.getStatusCode();
+
+		if (responseStatusCode == HttpStatus.UNAUTHORIZED) {
```
nenaraab: Wrong error handling: if you like to handle erronous calls you need to catch 
org.springframework.web.client.HttpClientErrorException

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlowTests.java
```diff
@@ -0,0 +1,237 @@
...
+		tokenFlow.client(clientId)
+				.secret(clientSecret)
+				.execute();
+
+		restTemplateMock.validateCallstate();
+		tokenDecoderMock.validateCallstate();
+	}
+
+	@Test
+	public void test_execute_throwsIfHttpStatusUnauthorized() {
```
nenaraab: need to fix this test as an unauthorized is represented by a HttpClientErrorException

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/RefreshTokenFlowTests.java
```diff
@@ -0,0 +1,223 @@
...
+
+		HttpEntity<Void> expectedRequest = buildExpectedRequest(clientId, clientSecret);
+
+		URI expectedRequestURI = UriComponentsBuilder.fromUri(TestConstants.tokenEndpointUri)
+				.queryParam("grant_type", "refresh_token")
+				.queryParam("refresh_token", refreshToken)
+				.build().toUri();
+
+		RestTemplateMock restTemplateMock = new RestTemplateMock(expectedRequestURI, expectedRequest, Map.class,
+				mockJwt.getTokenValue(), HttpStatus.UNAUTHORIZED);
```
nenaraab: need to fix this test as an unauthorized is represented by a HttpClientErrorException

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/UserTokenFlowTests.java
```diff
@@ -0,0 +1,301 @@
...
+		URI expectedRequestURI = UriComponentsBuilder.fromUri(TestConstants.tokenEndpointUri)
+				.queryParam("grant_type", "user_token")
+				.queryParam("response_type", "token")
+				.queryParam("client_id", clientId)
+				.build().toUri();
+
+		HttpEntity<Void> expectedRequest = buildExpectedRequest();
+
+		RestTemplateMock restTemplateMock = new RestTemplateMock(expectedRequestURI, expectedRequest, Map.class,
+				validMockJwt.getTokenValue(), HttpStatus.UNAUTHORIZED);
```
nenaraab: need to fix this test as an unauthorized is represented by a HttpClientErrorException

samples/spring-security-xsuaa-usage/README.md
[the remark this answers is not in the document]
nenaraab: Yes, if you have a look at SecurityContext.init() you will find out that we do exactly there what you describe in Readme and Demo :-)

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/RefreshTokenFlow.java
nenaraab: solved with `XsuaaDefaultEndpoints` and `OAuthServerEndpointsProvider`.

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/UserTokenFlow.java
nenaraab: solved with `XsuaaDefaultEndpoints` and `OAuthServerEndpointsProvider`.

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/XsuaaTokenFlowRequest.java
```diff
@@ -0,0 +1,130 @@
...
+		return this.oAuth2ServerEndpointsProvider.getTokenEndpoint();
+	}
+
+	/**
+	 * Returns the endpoint to authorize scopes. For example
+	 * {@code https://<server>:<port>/uaa/oauth/authorize}.
+	 *
+	 * @return the endpoint to authorize scopes.
+	 */
+	URI getAuthorizeEndpoint() {
```
nenaraab: authorizeEndpoint is never needed in context of token flows

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/flows/ClientCredentialsTokenFlow.java
nenaraab: return now the encoded token as String

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/flows/UserTokenFlowTests.java
nenaraab: Done

PR #134, opened 2019-09-06
Add a simple Base64 JWT decoder

abiskop:


abiskop:
@nenaraab - as discussed per e-mail

Issue #156, opened 2019-10-15
[token-client] Replace deprecated Date type


nenaraab:
LocalDateTime / or ZonedDateTime instead of Date?
https://github.wdf.sap.corp/MA/sdk/blob/develop/cloudplatform/cloudplatform-connectivity-scp-cf/src/main/java/com/sap/cloud/sdk/cloudplatform/connectivity/AccessToken.java

https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenResponse.java#L45

PR #161, opened 2019-10-16
Deprecated usage of java.util.Date, added java.time.Instant alternative

hassler-d:
I chose `Instant` in favour of `ZonedDateTime` (and `LocalDateTime`). Seemed to be the best fit. It is also used by spring in their OAuth2 Token implementation.


nenaraab:
closes #156

PR #171, opened 2019-10-22
Jwt decoder refactoring

hassler-d:
I have started refactoring the JwtDecoder part in spring-xsuaa. It is not finished yet but I just wanted to make sure that I am heading in the right direction. 

The main change is that `ReactiveXsuaaJwtDecoder/XsuaaJwtDecoder` now do not themselves create the NimbusJwtDecoder instances but receive a factory that is creating those decoders for them.
The `XsuaaJwtDecoderBuilder` uses default factories that deliver the decoder instances that were previously created inside `ReactiveXsuaaJwtDecoder/XsuaaJwtDecoder`. The main benefit with the approach is that a library user can create her own implementation of those factories. This makes the Decoder implementation interchangeable.


nenaraab:
Replaced by #173

PR #177, opened 2019-10-25
Disable default property source

hassler-d:
The default property source can now be disabled by setting the property `disable-default-property-source` to `true`.

SECAUTH-524

### DIFF
--- a/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java
+++ b/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java
@@ -3,10 +3,12 @@ package com.sap.cloud.security.xsuaa.autoconfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
 import org.springframework.security.oauth2.jwt.Jwt;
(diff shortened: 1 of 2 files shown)


spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java
```diff
@@ -57,6 +57,7 @@ public void configures_xsuaaRestTemplate() {
 	public void configures_xsuaaServiceConfiguration_withProperties() {
 		contextRunner
 				.withPropertyValues("spring.xsuaa.auto:true")
+				.withPropertyValues("spring.xsuaa.disable-default-property-source:false")
```
nenaraab: please remove this to check that by default property-source is configured :-)

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java
```diff
@@ -48,6 +50,20 @@ public XsuaaServiceConfiguration xsuaaServiceConfiguration() {
...
+	private static class PropertyConditions extends AllNestedConditions {
+
+		public PropertyConditions() {
+			super(ConfigurationPhase.PARSE_CONFIGURATION);
+		}
+
+		@ConditionalOnProperty(prefix = "spring.xsuaa", name = "multiple-bindings", havingValue = "false", matchIfMissing = true)
+		static class MultipleBindingsCondition { }
+
+		@ConditionalOnProperty(prefix = "spring.xsuaa", name = "disable-default-property-source", havingValue = "false", matchIfMissing = true)
```
nenaraab: matchIfMissing must be false i guess

hassler-d: The condition matches if the property is not set which means that the property source is **not** disabled (=enabled).
If it would be set to matchIfMissing = false then the property would explicitly be set to false to not disable the property source.

It works just like the _multiple-bindings_ property. If it is set to true then the default property source is disabled. In any other case it is enabled.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java
hassler-d: That is why it is set to false. False means it is **not disabeld**.

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java
nenaraab: Correct :-) I've checked the full test case. The default test case works without any changes!

PR #182, opened 2019-10-29
version 2.3.0-SNAPSHOT

nenaraab:


nenaraab:
@j-denner thanks :-)

@hassler-d maybe you can check how much we benefit from project-reactor as this will might support submodules only... (no IT, no sample project)
see also here:
https://www.baeldung.com/maven-multi-module

Issue #184, opened 2019-11-05
[spring-xsuaa] NimbusJwtDecoderJwkSupport is deprecated


nenaraab:
can we replace it with by NimbusJwtDecoder to remain downward compatible...?

nenaraab:
https://www.gitmemory.com/issue/spring-projects/spring-boot/17071/499177239

nenaraab:
Done with version 2.6.1

PR #190, opened 2019-11-18
#184 replaced deprecated NimbusJwtDecoder constructor

hassler-d:


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java
```diff
@@ -132,10 +130,11 @@ private Jwt verifyWithOnlineKey(String token, String jku, String kid) {
...
 	// TODO extract into separate class / bean
 	private JwtDecoder getDecoder(String jku) {
-		NimbusJwtDecoderJwkSupport decoder = new NimbusJwtDecoderJwkSupport(jku);
-		Optional.ofNullable(restOperations).ifPresent(decoder::setRestOperations);
-		decoder.setJwtValidator(tokenValidators);
-		return decoder;
+		NimbusJwtDecoder.JwkSetUriJwtDecoderBuilder jwkSetUriJwtDecoderBuilder = NimbusJwtDecoder.withJwkSetUri(jku);
+		if (restOperations != null) {
+			jwkSetUriJwtDecoderBuilder.restOperations(restOperations);
+		}
```
nenaraab: setJwtValidator missing... 

there should be a test that makes sure, that all validators (timestamp, audience, ... ) that are specified with XsuaaJwtDecoderBuilder are called.

nenaraab: see comment above

nenaraab:
will fix this issue: https://github.com/SAP/cloud-security-xsuaa-integration/issues/184

PR #198, opened 2019-12-10
Browser compatibility of basic auth

ilijanavonderweth:


samples/spring-security-basic-auth/pom.xml
```diff
@@ -32,7 +32,7 @@
 		<dependency>
 			<groupId>com.sap.cloud.security.xsuaa</groupId>
 			<artifactId>xsuaa-spring-boot-starter</artifactId>
-			<version>${project.version}</version>
+			<version>RELEASE</version>
```
nenaraab: please undo this change

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
```diff
@@ -63,4 +68,8 @@ BearerTokenResolver getTokenBrokerResolver() {
 		return converter;
 	}
 
+	@Bean
```
nenaraab: I guess you've already checked that, that this XsuaaServiceConfiguration bean is needed here!?

Do you know which condition is not met so that this [auto-configuration](https://github.com/SAP-samples/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfiguration.java#L43) is not considered in this sample?

If yes, I would appreciate a comment here in line 71 that explains why this is needed here :-)

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
```diff
@@ -30,14 +31,18 @@
 	@Override
 	protected void configure(HttpSecurity http) throws Exception {
 
+		//enforce browser login popup with basic authentication
+		BasicAuthenticationEntryPoint aep = new BasicAuthenticationEntryPoint();
```
nenaraab: pls rename aep to authEntryPoint or so...

PR #208, opened 2020-01-16
Migration

hassler-d:
Created migration guide and simplified migration in lib.

### DIFF
--- a/java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTestRule.java
+++ b/java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTestRule.java
@@ -29,6 +29,8 @@ import javax.servlet.Filter;
 import javax.servlet.Servlet;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
 import java.util.*;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.*;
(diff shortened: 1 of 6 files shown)


java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTestRule.java
```diff
@@ -158,6 +160,27 @@ public SecurityTestRule setPort(int port) {
 		return this;
 	}
 
+	/**
+	 * Same as {@link SecurityTestRule#setKeys(RSAKeys)} except you can pass in the
```
nenaraab: I would remove setKeys(RSAKey), its no longer needed.

java-security/src/main/java/com/sap/cloud/security/spring/adapter/SAPOfflineTokenServicesCloud.java
```diff
@@ -0,0 +1,59 @@
...
+		this(serviceConfiguration, () -> JwtValidatorBuilder.getInstance(serviceConfiguration).build());
+	}
+
+	SAPOfflineTokenServicesCloud(OAuth2ServiceConfiguration serviceConfiguration, Supplier<Validator<Token>> validatorSupplier) {
+		this.serviceConfiguration = serviceConfiguration;
+		this.validatorSupplier = validatorSupplier;
+	}
+
+	@Override
+	public OAuth2Authentication loadAuthentication(String accessToken)
```
nenaraab: is accessToken always != null?

-> have tests, when token is invalid (IllegalArgumentException)-> an invalidTokenException should be raised

search:
https://github.wdf.sap.corp/CPSecurity/java-container-security/search?q=InvalidTokenException&unscoped_q=InvalidTokenException

java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTestRule.java
hassler-d: You'r right, externally this is probably not needed anymore. Should we declare it  package protected since it is still used inside the test of the rule itself. Or would you rather use static keys for this test as well?

java-security/src/main/java/com/sap/cloud/security/spring/adapter/SAPOfflineTokenServicesCloud.java
hassler-d: I agree we should be more defensive here and check it!

java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTestRule.java
nenaraab: i would remove it and have only a test with the key files.

PR #231, opened 2020-02-21
Buildpack integration changes

hassler-d:
This PR are a collection of changes that were necessary to get the build pack integration working.

It does also contain a bug fix for [this method](https://github.com/SAP/cloud-security-xsuaa-integration/compare/BuildpackIntegration?expand=1#diff-6bcade52f5d1e4ee3532b801ab6c0a0eR108) which was not working. I added a method `toJsonString` to the `JsonObject` interface for that.
Another important change is in the `DefaultJsonObject.java` which was implemented with lazy data loading. (Meaning that it only parses the json string when the first method is called which actually reads some data).
This lead to a problem where exceptions were thrown inside (unrelated) validators complaining that the json is corrupt. This is very confusing. It is better to receive this exception as soon as the json object is created. 
I also created a utility class `FileReaderUtil` that reads file from disks (which `JwtGenerator` and `RSAKeys` use). With this it is possible to declare the path of the file inside resources which makes it easier and less error prone to write tests, e.g. [here](https://github.com/SAP/cloud-security-xsuaa-integration/compare/BuildpackIntegration?expand=1#diff-bb3463c258983b0f423ad8d05ed25765L50)

I move the first interface into the API project.

Last but not least this PR adds a new method to the JwtGenerator `withClaimsFromFile` which loads the claims defined in json file into the token that is being created.
I also added documentation.


java-security/src/main/java/com/sap/cloud/security/adapter/spring/SAPOfflineTokenServicesCloud.java
```diff
@@ -112,23 +113,19 @@ public SAPOfflineTokenServicesCloud(OAuth2ServiceConfiguration serviceConfigurat
 	@Override
 	public OAuth2Authentication loadAuthentication(@Nonnull String accessToken)
 			throws AuthenticationException, InvalidTokenException {
-		Token token = checkAndCreateToken(accessToken);
-
-		Set<String> scopes = token instanceof AccessToken
-				? ((AccessToken) token).getScopes()
-				: Collections.emptySet();
+		// TODO IAS support
+		AccessToken token = checkAndCreateToken(accessToken);
```
nenaraab: pls use Token instead of AccessToken

java-security/src/main/java/com/sap/cloud/security/adapter/spring/SAPOfflineTokenServicesCloud.java
```diff
@@ -159,7 +156,20 @@ public SAPOfflineTokenServicesCloud setLocalScopeAsAuthorities(boolean extractLo
...
+		clientAuthentication.setAuthorities(getAuthorities(scopes));
+		clientAuthentication.setApproved(true);
+		return clientAuthentication;
+	}
+
+	private Set<GrantedAuthority> getAuthorities(Collection<String> scopes) {
+		return scopes.stream().map(SimpleGrantedAuthority::new).collect(Collectors.toSet());
+	}
+
+	private XsuaaToken checkAndCreateToken(@Nonnull String accessToken) {
```
nenaraab: return type should be "Token" as well

java-security-test/src/main/java/com/sap/cloud/security/test/util/FileReaderUtil.java
```diff
@@ -0,0 +1,56 @@
...
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+/**
+ * Utility class that loads the content of files.
+ */
+public class FileReaderUtil {
```
nenaraab: why do we need that? 
Can't you use IOUtils.resourceToURL("/publicKey.txt").getPath() within the buildpack?

hassler-d: Yes you could do that of course. I just wanted to make it easier on the calling site. But I can remove this if you think it is not worth the extra class.

nenaraab: There is no use case and therefore no need. Thanks!

nenaraab:
@hassler-d solved merge conflicts in SapOfflineTokenServices

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/SpringOAuth2TokenKeyService.java
```diff
@@ -32,6 +32,9 @@ public String retrieveTokenKeys(URI tokenKeysEndpointUri) throws OAuth2ServiceEx
 		} catch (HttpClientErrorException ex) {
 			throw OAuth2ServiceException.createWithStatusCodeAndResponseBody("Error retrieving token keys",
 					ex.getStatusCode().value(), ex.getResponseBodyAsString());
+		} catch (Exception e) {
```
nenaraab: have you checked, whether this is also needed in DefaultOAuth2TokenKeyService?

PR #232, opened 2020-02-21
SpringSecurityContext for access within asynch threads

nenaraab:


java-security/src/main/java/com/sap/cloud/security/adapter/spring/SAPOfflineTokenServicesCloud.java
```diff
@@ -45,8 +51,20 @@
...
+ *     <ul>All Spring security features are supported that uses
+ *     {@link org.springframework.security.core.context.SecurityContext#getAuthentication()}</ul>
+ *     <ul>You can access the {@code Authentication} via {@link SecurityContextHolder#getContext()}
+ *     also within asynchronous threads.</ul>
+ *     <ul>You can access the {@code Token} via {@link SpringSecurityContext#getToken()}
+ *     also within asynchronous threads.</ul>
+ * </li>
+ *
  */
 public class SAPOfflineTokenServicesCloud implements ResourceServerTokenServices, InitializingBean {
```
hassler-d: The part below is changed in the build pack integration PR, which will lead to merge conflicts.

java-security/src/main/java/com/sap/cloud/security/adapter/spring/SpringSecurityContext.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     *
+     * @return the token or <code>null</code> if {@link SecurityContext} is empty
+     * or does not contain a token of this type.
+     */
+    @Nullable
+    public static Token getToken() {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        if(Objects.nonNull(authentication) && authentication.isAuthenticated() &&
+                authentication.getDetails() instanceof OAuth2AuthenticationDetails) {
```
hassler-d: Are you sure the token is inside the authDetails here? My probably have to check this.

java-security/src/test/java/com/sap/cloud/security/adapter/spring/SpringSecurityContextTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private static void setToken(Token token) {
+        SecurityContext context = new SecurityContextImpl();
+        OAuth2Authentication authentication = SAPOfflineTokenServicesCloud.getOAuth2Authentication(token,
+                "clientId", Collections.EMPTY_SET);
+
+        HttpServletRequest request = mock(HttpServletRequest.class);
+        when(request.getAttribute(OAuth2AuthenticationDetails.ACCESS_TOKEN_VALUE)).thenReturn(token.getTokenValue());
+
+        authentication.setDetails(new OAuth2AuthenticationDetails(request));
```
hassler-d: Ah I see here the details are set. We have to do this in SAPOfflineTokenServicesCloud as well then, right?

nenaraab: No, that is done by Spring Security Framework :-) magically

java-security/src/main/java/com/sap/cloud/security/adapter/spring/SpringSecurityContext.java
nenaraab: yes, I am. If it is not set for any reason, it will return null.

java-security/src/main/java/com/sap/cloud/security/adapter/spring/SAPOfflineTokenServicesCloud.java
nenaraab: yes, I'm aware of.
I've solved the conflict in the buildackintegration branch.

PR #235, opened 2020-03-03
Exented suppression rule because bug is still not fixed

hassler-d:
Extended suppression rule for 3 more months because https://github.com/jeremylong/DependencyCheck/issues/1827 is not fixed yet.


hassler-d:
Closed because I merged this PR into [this](https://github.com/SAP/cloud-security-xsuaa-integration/pull/237) one.

PR #249, opened 2020-03-18
Updated buildpack sample

hassler-d:
Can be merged as soon as this library is integrated in the SAP java buildpack


hassler-d:
Removed buildpack migration document from this branch and put it into its own branch here: https://github.com/SAP/cloud-security-xsuaa-integration/pull/262

PR #253, opened 2020-03-23
Cloud bulletinboard migration

hassler-d:
This PR adds two functionalities that made the migration of the [cloud-bulletinboard-ads](https://github.com/nenaraab/cloud-bulletinboard-ads) application easier. 

### withLocalScopes for JwtGenerator

  This makes it possible to set local scopes on the `JwtGenerator`. For this to work the `JwtGenerator`   needs to know which `appId` to use. I have solved this by adding the `withAppId` method to the   `JwtGenerator`. This is debatable and could also be added to the factory method of the   `JwtGenerator`. I chose the method because it is only needed when `withLocalScopes` is used.  `withLocalScopes` behaves like `withScopes` in the regard that consecutive calls always override the  
 old data.  This means if `withScopes("c", "d")` is called right after `withScopes("a", "b")` the resulting   token will only contain scopes ["c", "d"]. The same is true for `withLocalScopes`...
   **EDIT**: after discussion it overrides local scopes but not the (global) scopes provided with `withScopes("c", "d")`, see [comment](https://github.com/SAP/cloud-security-xsuaa-integration/pull/253#issuecomment-603235279) 

### Offline key verification for spring-xsuaa

  The token validator now falls back to the verification key of `VCAP_SERVICES` for the token signature 
  validation. This is basically the same behaviour like in java-security.


hassler-d:
After discussion: `withLocalScopes` and `withScopes` do not overwrite each other anymore. This is implemented in commit [3abfe765f7f3bdacb388527b0dd446f28708350d](https://github.com/SAP/cloud-security-xsuaa-integration/pull/253/commits/3abfe765f7f3bdacb388527b0dd446f28708350d)

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -224,13 +232,59 @@ public JwtGenerator withPrivateKey(PrivateKey privateKey) {
...
 		} else {
 			throw new UnsupportedOperationException("Scopes are not supported for service " + service);
 		}
 		return this;
 	}
 
+	/**
+	 * Works like {@link #withScopes(String...)}} but prefixes the scopes with
+	 * "appId.". For example if the appId is "xsapp", the scope "Read" will be
+	 * converted to "xsapp.Read". Make sure the appId has been via
```
nenaraab: ... provided / set

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -45,6 +46,9 @@
 
 	private JwtSignatureAlgorithm signatureAlgorithm;
 	private PrivateKey privateKey;
+	private String appId;
```
nenaraab: add comment // this is specific to XSUAA service

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -255,6 +309,12 @@ public Token createToken() {
 		}
 	}
 
+	private void putScopesInJsonPayload() {
+		List<String> resultingScopes = Stream.concat(localScopes.stream(), scopes.stream())
+				.collect(Collectors.toList());
+		jsonPayload.put(TokenClaims.XSUAA.SCOPES, resultingScopes);
```
nenaraab: can you replace
jsonPayload.put(TokenClaims.XSUAA.SCOPES, resultingScopes);
by
withClaimValues(TokenClaims.XSUAA.SCOPES, resultingScopes);
?

hassler-d:
@nenaraab 
In regard to the versioning of the library dependency of the samples there are two different approaches:

1. Samples that use a variable like java-security-usage: https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/samples/java-security-usage/pom.xml#L31

2. Others use the `${project.version}` since it is (by convention) set to the same value as the library version: https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/samples/spring-security-basic-auth/pom.xml#L35

Personally I would prefer option 1. because the samples are for library users and they might get confused that the library version is set  `${project.version}` like in option 2. which does not work unless they set their application version to the same value as the library.
They are more likely to create an explicit version variable like in option 1.

If we take option 1. I would add a version variable to all the samples and name it `sap.cloud.security.version`.

nenaraab:
hi @hassler-d 
prefer your prefered option as well. Is it okay, when i merge now? and you do the harmonization in another branch, as I'm done with code review :-)

PR #250, opened 2020-03-18
adding JUnit 5 support to java-security-test with usage examples

dirkrichtsteiger:
- a SecurityTestInitializer that behaves like the SecurityTestRule
- an example test at sample.spring.xsuaa.junitjupiter

PR #259, opened 2020-03-30
adding JUnit 5 support to java-security-test with usage examples

nenaraab:
- a SecurityTestInitializer that behaves like the SecurityTestRule
- an example test at sample.spring.xsuaa.junitjupiter


nenaraab:
@dirkrichtsteiger this closes #250 
usage is documented in your unit test :-)

java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTest.java
```diff
@@ -0,0 +1,348 @@
...
+    protected final List<FilterHolder> applicationServletFilters = new ArrayList<>();
+    // app server
+    protected Server applicationServer;
+    protected ApplicationServerOptions applicationServerOptions;
+    protected boolean useApplicationServer;
+
+    // mock server
+    protected WireMockServer wireMockServer;
+    protected RSAKeys keys;
+    protected int wireMockPort = 0;
```
hassler-d: I think we could get rid of that instance variable now because it is not really needed anymore.

nenaraab: should be done

PR #262, opened 2020-04-06
Buildpack migration guide

hassler-d:
This PR contains the migration guide for buildpack applications. It will be relevant as soon as this library is integrated in the buildpack.

### DIFF
--- a/java-security/Migration_SAPJavaBuildpackProjects.md
+++ b/java-security/Migration_SAPJavaBuildpackProjects.md
@@ -1,53 +1,239 @@
 # Migration Guide for Applications that use SAP Java Buildpack and java-container-security
 
-This migration guide is a step-by-step guide explaining how to replace the following SAP-internal Java Container Security Client libraries
-- com.sap.xs2.security:java-container-security
-- com.sap.cloud.security.xsuaa:java-container-security  
-
-when using **SAP Java Buildpack**.
+**This document is only applicable as soon as the integration of this security library into the buildpack as been finished!**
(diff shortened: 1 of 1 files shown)


java-security/Migration_SAPJavaBuildpackProjects.md
```diff
@@ -1,53 +1,239 @@
...
+
+Now you will most likely have compile error at places where the old api module was used. Those are the parts of the
+application where information from the token (or user info) was read. See the next sections on how to migrate that.
+
+## User principal <a name="user-principal"></a>
+
+In the old version of the buildpack the user principal could be obtained via:
+
+```java
+XSUserInfo userInfo = (XSUserInfo) request.getUserPrincipal();
```
nenaraab, line 69: Is this possible again? With revert of UserInfoHolder...?

hassler-d: No, not in the current state of the integration. The principal does not implement both interfaces (`XSUserInfo` and `AccessToken`).
What is possible now is that the user loads the `UserInfo` from the `UserInfoHolder`.

nenaraab: okay, but then there is no benefit from using the XSUserInfoAdapter...

hassler-d: We need the adapter to support XSUserHolder (otherwise would we need the old library).
Additionally the adapter enables an easier transition for the users.

PR #267, opened 2020-04-20
User info compatibility

hassler-d:
This PR improves UserInfo compatibility. I have added an integration test that consists of the original UserInfo test and its test resources.
I have also synchronised the deprecation annotations on the XSUserInfo interface. 

I just noticed that there is still one thing be missing when using the `requestToken` method of `UserInfo`: We do not provide an implementation of `XSTokenRequest`. Callers of this method would need to provide their own implementation.

Should we provide an implementation in the library? If so where and should this also be part of this PR?


java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -318,16 +357,77 @@ private String getClaimValue(String claimname) throws XSUserInfoException {
...
+		try {
+			ccfToken = xsuaaTokenFlows.clientCredentialsTokenFlow()
+					.subdomain(getSubdomain())
+					.attributes(additionalAuthAttributes)
+					.execute().getAccessToken();
+		} catch (TokenFlowException e) {
+			throw new XSUserInfoException("Error performing Client Credentials Flow. See exception cause.", e);
+		}
+
+		LOGGER.info("Got the Client Credentials Flow Token: {}", ccfToken);
```
nenaraab, line 428: never log tokens!

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -318,16 +357,77 @@ private String getClaimValue(String claimname) throws XSUserInfoException {
...
+					.subdomain(getSubdomain())
+					.token(getAppToken())
+					.attributes(additionalAuthAttributes)
+					.execute().getAccessToken();
+		} catch (TokenFlowException e) {
+			throw new XSUserInfoException("Error performing User Token Flow. See exception cause.", e);
+		}
+
+		LOGGER.info("Got the exchanged token for 3rd party service (clientId: {}) : {}", clientId, userToken);
+		LOGGER.info("You can now call the 3rd party service passing the exchanged token value: {}. ", userToken);
```
nenaraab, line 411: never log tokens!

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -34,40 +48,52 @@
...
-	public XSUserInfoAdapter(Token accessToken) throws XSUserInfoException {
-		this(accessToken, Environments.getCurrent().getXsuaaConfiguration());
+	public XSUserInfoAdapter(Object accessToken) {
+		this(accessToken, Environments.getCurrent().getXsuaaConfiguration(), new DefaultOAuth2TokenService());
 	}
 
-	public XSUserInfoAdapter(AccessToken accessToken) throws XSUserInfoException {
-		this(accessToken, null);
+	public XSUserInfoAdapter(AccessToken accessToken) {
+		this(accessToken, Environments.getCurrent().getXsuaaConfiguration(), new DefaultOAuth2TokenService());
```
nenaraab, line 63: this is incompatible: new DefaultOAuth2TokenService() requires apache rest client, but this is only "provided"... there might be other projects, that like to use Spring RestTemplate 

Question: is apache rest client provided as part of the SAP buildpack?

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -318,16 +357,77 @@ private String getClaimValue(String claimname) throws XSUserInfoException {
...
-	String getExternalAttribute(String attributeName) throws XSUserInfoException {
+	String getExternalAttribute(String attributeName) {
 		return getAttributeFromClaimAsString(EXTERNAL_ATTRIBUTE, attributeName);
 	}
 
+	private String performTokenFlow(String baseUrl, int tokenRequestType, String clientId, String clientSecret,
+			Map<String, String> additionalAuthAttributes) {
+		try {
+			ClientCredentials clientCredentials = new ClientCredentials(clientId, clientSecret);
+			XsuaaTokenFlows xsuaaTokenFlows = new XsuaaTokenFlows(oAuth2TokenService,
```
nenaraab, line 376: raise UnsupportedOperationException in case oAuth2TokenService is null... furthermore allow a setter

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -34,40 +48,52 @@
...
+	XSUserInfoAdapter(Object accessToken, OAuth2ServiceConfiguration configuration,
+			OAuth2TokenService oAuth2TokenService) {
 		if (!(accessToken instanceof AccessToken)) {
 			String type = Objects.isNull(accessToken) ? null : accessToken.getClass().getName();
 			throw new XSUserInfoException("token is of instance " + type
 					+ " but needs to be an instance of AccessToken.");
 		}
 		this.accessToken = (AccessToken) accessToken;
 		this.configuration = configuration;
+		this.oAuth2TokenService = oAuth2TokenService;
```
nenaraab, line 75: Is it possible to check, if apache rest client is on class path => create DefaultOAuth2TokenService() or create SpringOAuth2TokenService()

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
hassler-d: Okay, removed!

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
hassler-d: The apache rest client is provided internally by the same class loader like `XSUserInfoAdapter`. It is not on the class path of the application however. I am not 100% sure if this is a problem and will check if that works in a sample/arquillian test. 

If someone uses the XSUserInfoAdapter standalone **without** buildpack, then they would need to bring their own apache rest client.

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
hassler-d: Alright could make sense to override that for some applications and then we cannot control and should check for `null`, I will do that.

java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
hassler-d: For buildpack users that just works now.

PR #272, opened 2020-04-28
Sample and Documentation: IAS identity kernel service

nenaraab:
### Overview
This sample is a Java back-end application running on the Cloud Foundry. For all incoming requests it checks whether the IAS token is valid. 
Technically, the user is authenticated using the 
[`IasTokenAuthenticator`](/java-security/src/main/java/com/sap/cloud/security/servlet) which is defined in the [Java Security](../../java-security/) library.

### 2020-20-12
The name of the identity service will be "identity" and the plan name is "application".


hassler-d (approved):
LGTM! Only small comments

PR #273, opened 2020-04-29
Improved IAS logging

hassler-d:
This PR prevents error logs when IAS is not actually used. It also adds a test and improves the log in `SecurityContext`.

### DIFF
--- a/java-api/src/main/java/com/sap/cloud/security/config/Service.java
+++ b/java-api/src/main/java/com/sap/cloud/security/config/Service.java
@@ -28,7 +28,7 @@ public enum Service {
 	 */
 	public String getCFName() {
 		if (this == IAS && cloudFoundryName == null) {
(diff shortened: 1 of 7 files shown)


java-api/src/main/java/com/sap/cloud/security/config/Service.java
```diff
@@ -29,7 +28,7 @@ private static String getIasServiceName() {
 	 */
 	public String getCFName() {
 		if (this == IAS && cloudFoundryName == null) {
-			LoggerFactory.getLogger(Service.class).error("IAS Service is not yet supported!!!"); // TODO 2.6.2 remove
+			LoggerFactory.getLogger(Service.class).info("IAS Service is not yet supported!!!"); // TODO 2.6.2 remove
```
nenaraab, line 31: this should remain errror

hassler-d: okay

Issue #277, opened 2020-05-08
nullpointerexception


scorvio:
Caused by: java.lang.NullPointerException: null
	at com.sap.cloud.security.xsuaa.XsuaaServicesParser.getJSONObjectFromTag(XsuaaServicesParser.java:143) ~[spring-xsuaa-2.6.2.jar:na]
	at com.sap.cloud.security.xsuaa.XsuaaServicesParser.searchXsuaaBinding(XsuaaServicesParser.java:128) ~[spring-xsuaa-2.6.2.jar:na]
	at com.sap.cloud.security.xsuaa.XsuaaServicesParser.parseCredentials(XsuaaServicesParser.java:114) ~[spring-xsuaa-2.6.2.jar:na]
	at com.sap.cloud.security.xsuaa.XsuaaServicesParser.parseCredentials(XsuaaServicesParser.java:96) ~[spring-xsuaa-2.6.2.jar:na]
	at com.sap.cloud.security.xsuaa.XsuaaServicePropertySourceFactory.createPropertySource(XsuaaServicePropertySourceFactory.java:60) ~[spring-xsuaa-2.6.2.jar:na]
	at org.springframework.context.annotation.ConfigurationClassParser.processPropertySource(ConfigurationClassParser.java:455) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassParser.doProcessConfigurationClass(ConfigurationClassParser.java:274) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassParser.processConfigurationClass(ConfigurationClassParser.java:245) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassParser.parse(ConfigurationClassParser.java:194) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassParser.doProcessConfigurationClass(ConfigurationClassParser.java:298) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassParser.processConfigurationClass(ConfigurationClassParser.java:245) ~[spring-context-5.2.0.RELEASE.jar:5.2.0.RELEASE]
[...]

PR #278, opened 2020-05-08
BugFix: NullpointerException when VCAPServices provides no 'plan' 

nenaraab:
### Related Github issue
#277


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaServicesParser.java
```diff
@@ -139,9 +139,9 @@ private static JSONObject getJSONObjectFromTag(final JSONArray jsonArray, String
 			JSONObject binding = (JSONObject) jsonArray.get(i);
 			JSONArray tags = (JSONArray) binding.get(TAGS);
 
+			Optional<String> planName = Optional.ofNullable(binding.getAsString("plan"));
+			boolean isApiAccessPlan = (planName.isPresent() && planName.get().equals("apiaccess")) ? true : false;
```
hassler-d, line 143: You could remove `? true : false` here and just use the result of the expression before the `?` operator :)

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/extractor/LocalAuthoritiesExtractorTest.java
```diff
@@ -0,0 +1,55 @@
...
+        scopes.add("appId2!888.Scope1");
+        scopes.add("appId2!777.Scope3");
+        Mockito.when(token.getScopes()).thenReturn(scopes);
+    }
+
+    @Test
+    public void extractLocalScopes() {
+        assertThat(cut.getAuthorities(token)).hasSize(2);
+        assertThat(cut.getAuthorities(token)).contains(new SimpleGrantedAuthority("Scope1"));
+        assertThat(cut.getAuthorities(token)).contains(new SimpleGrantedAuthority("Scope2"));
```
hassler-d, line 36: That could be written with a single assert like: 
`assertThat(cut.getAuthorities(token)).containsExactlyInAnyOrder(new SimpleGrantedAuthority("Scope1"), new SimpleGrantedAuthority("Scope2"))`

hassler-d: Same goes for the other test cases :)

PR #280, opened 2020-05-12
Token client cache

hassler-d:
This PR adds a Cache to the token client. The cache can be configured and disabled for the token service.

### DIFF
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
@@ -27,15 +27,17 @@ public abstract class AbstractOAuth2TokenService implements OAuth2TokenService, 
 	private final Ticker ticker; // used for cache testing
 
 	public AbstractOAuth2TokenService() {
-		 this(Ticker.systemTicker(), false);
+		this(Ticker.systemTicker(), false);
 	}
 
 	/**
 	 * This constructor is used for testing purposes only.
 	 *
-	 * @param cacheTicker will be used in the cache to determine the time.
-	 * @param sameThreadCache set to true disables maintenance jobs of the cache. This makes
-	 *                        the cache slower but more predictable for testing.
+	 * @param cacheTicker
+	 *            will be used in the cache to determine the time.
+	 * @param sameThreadCache
+	 *            set to true disables maintenance jobs of the cache. This makes the
+	 *            cache slower but more predictable for testing.
 	 */
 	AbstractOAuth2TokenService(Ticker cacheTicker, boolean sameThreadCache) {
 		ticker = cacheTicker;
@@ -65,7 +67,6 @@ public abstract class AbstractOAuth2TokenService implements OAuth2TokenService, 
 
 		return getOAuth2TokenResponse(tokenEndpointUri, headers, parameters, subdomain);
 	}
-
 
 	@Override
 	public OAuth2TokenResponse retrieveAccessTokenViaUserTokenGrant(@Nonnull URI tokenEndpointUri,
@@ -172,12 +173,12 @@ public abstract class AbstractOAuth2TokenService implements OAuth2TokenService, 
 			@Nullable String subdomain) throws OAuth2ServiceException {
 		URI tokenEndpointUriWithSubdomainReplaced = UriUtil.replaceSubdomain(tokenEndpointUri, subdomain);
 		if (isCacheEnabled()) {
-			return getOrRequestResponse(headers, tokenEndpointUriWithSubdomainReplaced, additionalParameters);
+			return getOrRequestAccessToken(headers, tokenEndpointUriWithSubdomainReplaced, additionalParameters);
 		}
 		return requestAccessToken(tokenEndpointUriWithSubdomainReplaced, headers, additionalParameters);
 	}
 
-	private OAuth2TokenResponse getOrRequestResponse(HttpHeaders headers, URI tokenEndpointUriWithSubdomainReplaced,
+	private OAuth2TokenResponse getOrRequestAccessToken(HttpHeaders headers, URI tokenEndpointUriWithSubdomainReplaced,
 			Map<String, String> additionalParameters) throws OAuth2ServiceException {
 		CacheKey cacheKey = new CacheKey(tokenEndpointUriWithSubdomainReplaced, headers, additionalParameters);
 		OAuth2TokenResponse oAuth2TokenResponse = getOrCreateResponseCache().getIfPresent(cacheKey);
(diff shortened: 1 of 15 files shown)


token-client/src/main/java/com/sap/cloud/security/xsuaa/tokenflows/CacheConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+
+	/**
+	 * Special cache configuration used to disable caching.
+	 */
+	public static final CacheConfiguration CACHE_DISABLED = new CacheConfiguration(Duration.ZERO, 0);
+
+	private final Duration expireAfterWrite;
+	private final int cacheSize;
+
+	public CacheConfiguration(Duration expireAfterWrite, int cacheSize) {
```
nenaraab, line 24: i would prefer a private Constructor and a getInstance() method

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
```diff
@@ -1,21 +1,52 @@
...
+	 *            cache slower but more predictable for testing.
+	 */
+	AbstractOAuth2TokenService(Ticker cacheTicker, boolean sameThreadCache) {
+		ticker = cacheTicker;
+		this.sameThreadCache = sameThreadCache;
+	}
+
+	@Override
+	public void clearCache() {
+		getOrCreateResponseCache().invalidateAll();
```
nenaraab, line 49: cache creation seems not to be required...?

if (Objects.nonNull(responseCache)) { responseCache.invalidateAll(); }

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
```diff
@@ -137,4 +168,78 @@ public OAuth2TokenResponse retrieveAccessTokenViaJwtBearerTokenGrant(URI tokenEn
 	protected abstract OAuth2TokenResponse requestAccessToken(URI tokenEndpointUri, HttpHeaders headers,
 			Map<String, String> parameters) throws OAuth2ServiceException;
 
+	private OAuth2TokenResponse getOAuth2TokenResponse(@Nonnull URI tokenEndpointUri, HttpHeaders headers,
+			Map<String, String> additionalParameters,
```
nenaraab, line 172: pls rename additionalParameters to parameters

token-client/src/main/java/com/sap/cloud/security/xsuaa/http/HttpHeaders.java
```diff
@@ -17,4 +18,19 @@ public HttpHeaders(List<HttpHeader> headers) {
...
 	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o)
+			return true;
+		if (o == null || getClass() != o.getClass())
+			return false;
+		HttpHeaders that = (HttpHeaders) o;
+		return Objects.equals(headers, that.headers);
```
nenaraab, line 29: what about the order of the headers? Is it guaranteed, that they are always added in the same order?

token-client/src/main/java/com/sap/cloud/security/xsuaa/tokenflows/CacheConfiguration.java
hassler-d: Done :)

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
hassler-d: I have reworked this part of the code, please have a look at it again.

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
```diff
@@ -137,4 +168,78 @@ public OAuth2TokenResponse retrieveAccessTokenViaJwtBearerTokenGrant(URI tokenEn
...
+			@Nullable String subdomain) throws OAuth2ServiceException {
+		URI tokenEndpointUriWithSubdomainReplaced = UriUtil.replaceSubdomain(tokenEndpointUri, subdomain);
+		if (isCacheEnabled()) {
+			return getOrRequestAccessToken(headers, tokenEndpointUriWithSubdomainReplaced, additionalParameters);
+		}
+		return requestAccessToken(tokenEndpointUriWithSubdomainReplaced, headers, additionalParameters);
+	}
+
+	private OAuth2TokenResponse getOrRequestAccessToken(HttpHeaders headers, URI tokenEndpointUriWithSubdomainReplaced,
```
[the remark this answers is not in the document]
hassler-d: done!

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
hassler-d: done!

token-client/src/main/java/com/sap/cloud/security/xsuaa/http/HttpHeaders.java
hassler-d: For the equality of `HttpHeaders` right now the order **does** matter. In my opinion the order should not matter and I would rather change it so it doesn't!

However, for the token client this does not really matter since we control the creation of the HttpHeaders 100% and for every request type the `HttHeaders` the same.

It would still prefer to make them robust against order changes. That is a small change + test. Can I add it in this PR?

nenaraab: yes, pls :-) thanks!

hassler-d: Done :)

hassler-d:
> One thing I'm missing here, is the check whether the token is expired in the next seconds and in this case not to return the cached one.

Should I add this here or in a new PR?

nenaraab:
entweder

> > One thing I'm missing here, is the check whether the token is expired in the next seconds and in this case not to return the cached one
> 
> @nenaraab Should I add those requirements here or in a new PR?

pls add it to this PR or if its easier to PR #281

nenaraab:
@hassler-d 

please adapt the configuration accordingly:
- reduce duration to 10 minutes
- increase size to 1000

hassler-d:
> @hassler-d
> 
> please adapt the configuration accordingly:
> 
> * reduce duration to 10 minutes
> * increase size to 1000

done!

PR #284, opened 2020-05-12
TG.02: Basic support of Zones (token claim "zone_uuid")

nenaraab:
Use getSubaccountId() only to fetch the subaccount id, e.g. for calling the metering API for user-based pricing.
In case you are interested in the customers tenant GUID make use of getZoneId() instead!
In upcoming releases - especially for new subaccounts - subaccount id will no longer match the tenant GUID which is provided via the xsuaa access token as zid claim or via the ias oidc token as zone_uuid claim.

**Enhance java-security:**

- new method Tenant.getZoneId() should return "zone_uuid" if available or as fallback "zid" for xsuaa tokens

**Enhance spring-xsuaa** - no incompatible change fur application developers

- deprecate XSUserInfo.getIdentityZone() currently returns "zid" => still ok
- add XSUserInfo.getZoneId() which should return "zid" for xsuaa tokens
- update **Token.getSubaccountId()** currently returns "zid" => should return "subaccountid" if available or as fallback "zid". 
- documents that **Token.getZoneId()** should be used to get tenant discriminator (tenant guid). For most of the old subaccounts this matches the subaccount id.


.github/workflows/maven.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           java-version: 1.8
       - name: Build with Maven
-        run: mvn -B package --file pom.xml
+        run: mvn -B install --file pom.xml
```
hassler-d, line 21: The build failed because it contains changes used by the samples. The samples retrieve their dependencies via maven central and therefore they fail to build.
This change makes the build succeed because the dependencies are being installed locally.

@nenaraab Are the dependencies not installed in the build environment on purpose? If so we have to revert this commit and it will only work build again after a release.

nenaraab: No, i think this should be ok and will help us a lot to keep our pipeline running, even after version update 👍

api/src/main/java/com/sap/xsa/security/container/XSUserInfo.java
```diff
@@ -78,36 +78,46 @@
...
-	 * Return subaccount identifier which is the same like the identity zone (tenant
-	 * id).
-	 * 
-	 * @deprecated Can be replaced with
-	 *             {@code token.getClaimAsString(TokenClaims.XSUAA.ZONE_ID)} from
-	 *             the {@code com.sap.cloud.security.token} package.
+	 * Return subaccount identifier which is in most cases same like the
+	 * identity zone.
+	 * DO only use this for metering purposes.
+	 * DO NOT longer use this method to get the unique tenant id"! For that use {@link #getZoneId()}.
```
hassler-d, line 95: I think the double quotes are a typo here?

PR #287, opened 2020-05-15
[java-security] SECAUTH-813: support more than 1 other configuration

nenaraab:
SAPOfflineTokenServicesCloud and JwtValidatorBuilder should more than 1 other configuration
- OAuth2ServiceConfiguration should implement equals method

TODO:
 - [ ] wait on stakeholder feedback


java-security/src/test/java/com/sap/cloud/security/token/validation/validators/JwtValidatorBuilderTest.java
```diff
@@ -110,4 +112,34 @@ public void buildWithAnotherValidator_containsAddedValidator() {
...
+				.configureAnotherServiceInstance(xsuaaConfig3)
+				.build();
+
+		for (Validator validator: combiningValidator.getValidators()) {
+			if(validator instanceof JwtAudienceValidator) {
+				assertThat(((JwtAudienceValidator) validator).clientIds).containsAll(clientIds);
+				return;
+			}
+		}
+		assertThat(true).isFalse(); // should never be called
```
hassler-d, line 142: I would prefer to let the test fail with a message like `Assert.fail("No JwtAudienceValidator found that contains all clientIds!");`

nenaraab:
@hassler-d 
thanks for your review. I've applied all of your review comments...
please perform another review :-)

Issue #292, opened 2020-05-27
[token-client] add more debug logs and details to exception


nenaraab:
1. provide further infos about the outgoing request:

```
"\t... 12 more","Caused by: com.sap.cloud.security.xsuaa.client.OAuth2ServiceException: Unexpected error retrieving JWT token: Error retrieving JWT token. Received status code 401. Call to OAuth2 server was not successful: {\"error\":\"unauthorized\",\"error_description\":\"Bad credentials\"}",
"\tat com.sap.cloud.security.xsuaa.client.DefaultOAuth2TokenService.executeRequest(DefaultOAuth2TokenService.java:59)",
"\tat com.sap.cloud.security.xsuaa.client.DefaultOAuth2TokenService.requestAccessToken(DefaultOAuth2TokenService.java:44)",
"\tat com.sap.cloud.security.xsuaa.client.AbstractOAuth2TokenService.retrieveAccessTokenViaClientCredentialsGrant(AbstractOAuth2TokenService.java:36)",
"\tat com.sap.cloud.security.xsuaa.tokenflows.ClientCredentialsTokenFlow.requestTechnicalUserToken(ClientCredentialsTokenFlow.java:133)",
"\t... 15 more"]

```
https://github.com/SAP/cloud-security-xsuaa-integration/blob/e5764b61c231979d8643ffa3b962ad95ff48fcf1/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2ServiceException.java#L16

2. add logs with debug level:
- when was a token fetched from cache
- when was a token fetched from cache, but expired
- when was a token requested from xsuaa

3. document how to increase log level for apache rest client and restTemplate in troubleshoot section
(https://hc.apache.org/httpcomponents-client-4.5.x/logging.html)
org.apache.http.wire: TraceLog.txt.

nenaraab:
merged in master and will be available with version 2.7.1

Issue #223, opened 2020-02-07
Supportability via SAP_EXT_TRL


andre68723:
It used to be possible to see the jwt token in the application log by providing env parameters 
```
cf/xs set-env <app> SAP_EXT_TRC stdout
cf/xs set-env <app> SAP_EXT_TRL 3
cf/xs restage <app>
```
This made life a lot easier when jwt tokens were needed to be obtained for support issues. 
Seems to come originally from SAPSSOEXT library

nenaraab:
As implemented here:
https://github.com/SAP-samples/cloud-sapjwt#error-situations

nenaraab:
@hassler-d 
Can you please take over?

hassler-d:
> @hassler-d
> Can you please take over?

Yes sure! 



> It used to be possible to see the jwt token in the application log by providing env parameters
> 
> ```
> cf/xs set-env <app> SAP_EXT_TRC stdout
> cf/xs set-env <app> SAP_EXT_TRL 3
> cf/xs restage <app>
> ```
> 
Can you explain to me or show me some information about what SAP_EXT_TRL means?

hassler-d:
Hello @andre68723,

we have extended the logging. If you set the logging level to debug and the validation of JWT tokens fails, you 
will see a log message like this:

```log
[main] WARN com.sap.cloud.security.token.validation.ValidationResults - Issuer is not trusted because 'iss' 'http://auth.com' does not match host 'myauth.com' of the identity provider.
[main] DEBUG com.sap.cloud.security.token.validation.CombiningValidator - Validator that caused the failed validation: com.sap.cloud.security.token.validation.validators.JwtIssuerValidator
[main] DEBUG com.sap.cloud.security.token.validation.CombiningValidator - Object of type com.sap.cloud.security.token.XsuaaToken that caused the failed validation: 
Jwt header
	{"jku":"http://auth.com/token_keys","kid":"key-id-0","alg":"RS256"}
Jwt payload
	{"aud":["clientId"],"ext_attr":{"enhancer":"XSUAA","zdn":"theSubdomain"},"grant_type":"user_token","user_name":"testUser","origin":"userIdp","iss":"http://auth.com","exp":"6974031600","cid":"clientId"}
```
This would display the content of JWT header and payload. This feature will be part of the 2.7.5 release of this library.

It will be available in the SAP Java Buildpack with the next release (by the end of next week).
[...]

PR #294, opened 2020-05-28
Improved logging

hassler-d:
This PR improves the logging so that the requirements of #292 are fulfilled.
It also adds a TokenLogger util class that lets you safely log token content for debugging purposes.

### DIFF
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
@@ -214,7 +214,8 @@ public abstract class AbstractOAuth2TokenService implements OAuth2TokenService, 
 
 	private OAuth2TokenResponse getOrRequestAccessToken(URI tokenEndpoint, HttpHeaders headers,
 			Map<String, String> parameters) throws OAuth2ServiceException {
-		LOGGER.debug("Getting token for token endpoint uri={} with headers={} and parameters={}", tokenEndpoint, headers, parameters);
+		LOGGER.debug("Getting token for token endpoint uri={} with headers={} and parameters={}", tokenEndpoint,
+				headers, parameters);
 		CacheKey cacheKey = new CacheKey(tokenEndpoint, headers, parameters);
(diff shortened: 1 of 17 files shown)


hassler-d:
@nenaraab Should #223 be included into this PR as well?

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyService.java
```diff
@@ -36,8 +36,10 @@ public String retrieveTokenKeys(URI tokenKeysEndpointUri) throws OAuth2ServiceEx
 			if (statusCode == HttpStatus.SC_OK) {
 				return bodyAsString;
 			} else {
-				throw OAuth2ServiceException
-						.createWithStatusCodeAndResponseBody("Error retrieving token keys", statusCode, bodyAsString);
+				throw OAuth2ServiceException.builder("Error retrieving token keys")
+						.withStatusCode(statusCode)
```
nenaraab, line 40: i would have expected 
.withUri(tokenKeysEndpointUri)

hassler-d: Thanks I missed that one!

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenService.java
```diff
@@ -207,20 +217,38 @@ private OAuth2TokenResponse getOAuth2TokenResponse(@Nonnull URI tokenEndpointUri
...
+	}
+
+	private void logDebug(OAuth2TokenResponse response) {
+		String accessToken = response.getAccessToken();
+		String refreshToken = response.getRefreshToken();
+		if (accessToken != null) {
+			tokenLogger.logDebug(accessToken, "Access token:");
+		}
+		if (refreshToken != null) {
+			tokenLogger.logDebug(refreshToken, "Refresh token:");
```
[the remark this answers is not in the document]
hassler-d: Yes, thanks! I just saw that the refresh_token is in another format so it can't even be logged with the `TokenLogger` Utility! It only works with JWT tokens. I will remove that part then!

PR #296, opened 2020-05-29
Token key cache configurable from XsuaaTokenAuthenticator

hassler-d:
This PR adds a new interface `CacheConfiguration` to the `java-api` package that both `TokenKeyCache` and `TokenCache` implement.
It is now possible to configure the `XsuaaTokenAuthenticator` with a cache configuration for the token key cache.

I also added `isCacheDisabled()` to the CacheConfiguration interface that implementations should use to check if the cache is disabled. I made the `DEFAULT` and `CACHE_DISABLED` constants private and accesible with `TokenCacheConfiguration.defaultConfiguration()` and `TokenCacheConfiguration.cacheDisabled()`. 

It is still not possible to disable the token key cache.


java-security/pom.xml
```diff
@@ -33,10 +33,6 @@
 			<artifactId>token-client</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-		<dependency>
```
nenaraab, line 36: great fix :-) did not recognize

nenaraab:
Hi @hassler-d 
somehow it has conflicts with master branch.
The rest is ready for merge!!!
Thanks a lot!

PR #300, opened 2020-06-03
do not expose TokenKeyCacheConfiguration, set classes to package private

nenaraab:
@hassler-d pls. check whether package private is okay for these classes:
- TokenKeyCacheConfiguration (new)
- OAuth2TokenKeyServiceWithCache
- all Validators, as the JwtValidatorBuilder should be used instead as documented in README.md

Furthermore - as discussed - I've removed the header from the token cache.


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/OAuth2TokenKeyServiceWithCache.java
```diff
@@ -29,7 +28,7 @@
  * Decorates {@link OAuth2TokenKeyService} with a cache, which gets looked up
  * before the identity service is requested via http.
  */
-public class OAuth2TokenKeyServiceWithCache implements Cacheable {
+class OAuth2TokenKeyServiceWithCache implements Cacheable {
```
hassler-d, line 31: This class has been public since November. For the cache configuration in the buildpack I do not need it. Were other People using this class? Probably not. I think we can move this to package private.

nenaraab:
@hassler-d can you please check my last commit as well... thx

nenaraab:
thanks, @hassler-d for double check!

PR #306, opened 2020-06-16
[java-security] derive audiences from scopes

nenaraab:
this is relevant for apps that migrate from java-security-container client lib and were called with a token of grant type "user_token", which does not provide `aud` claim.

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
@@ -3,6 +3,7 @@ package com.sap.cloud.security.token.validation.validators;
 import static com.sap.cloud.security.xsuaa.Assertions.assertHasText;
 
 import com.sap.cloud.security.token.Token;
+import com.sap.cloud.security.token.TokenClaims;
 import com.sap.cloud.security.token.validation.ValidationResult;
 import com.sap.cloud.security.token.validation.ValidationResults;
 import com.sap.cloud.security.token.validation.Validator;
@@ -10,9 +11,7 @@ import com.sap.cloud.security.token.validation.Validator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.LinkedHashSet;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 
 /**
  * Validates if the jwt access token is intended for the OAuth2 client of this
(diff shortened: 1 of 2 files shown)


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
```diff
@@ -94,6 +93,15 @@ private ValidationResult validateAudienceOfXsuaaBrokerClone(Set<String> allowedA
 				audiences.add(audience);
 			}
 		}
+		// extract audience (app-id) from scopes
+		if (audiences.size() == 0) {
```
hassler-d, line 97: It would be more idiomatic to use `isEmpty` here

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
```diff
@@ -94,6 +93,15 @@ private ValidationResult validateAudienceOfXsuaaBrokerClone(Set<String> allowedA
...
+			for (String scope : token.getClaimAsStringList(TokenClaims.XSUAA.SCOPES)) {
+				if (scope.contains(".")) {
+					String aud = scope.substring(0, scope.indexOf('.'));
```
hassler-d, line 100: This looks a lot like the code in line 88 but without calling `trim`.
Is `trim` not needed here? If it is I would suggest to extract that into a helper that strips the namespace. This can be used here and on line 89 then.

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
```diff
@@ -94,6 +93,15 @@ private ValidationResult validateAudienceOfXsuaaBrokerClone(Set<String> allowedA
 				audiences.add(audience);
 			}
 		}
+		// extract audience (app-id) from scopes
+		if (audiences.size() == 0) {
+			for (String scope : token.getClaimAsStringList(TokenClaims.XSUAA.SCOPES)) {
+				if (scope.contains(".")) {
```
hassler-d, line 99: I would suggest to use the constant DOT here because it is already defined in the class.

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
nenaraab: Done

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
nenaraab: Done

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java
nenaraab: Done

nenaraab:
@hassler-d 
thanks for your feedback!
I've incorporated them.

nenaraab:
@hassler-d can you please perform a final review?

PR #309, opened 2020-06-19
refactor jku validation

nenaraab:
- increased test coverage

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
@@ -4,7 +4,6 @@ import static com.sap.cloud.security.token.validation.ValidationResults.*;
 import static com.sap.cloud.security.xsuaa.Assertions.assertHasText;
 
 import java.net.URI;
-import java.net.URISyntaxException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
```diff
@@ -38,27 +37,37 @@
...
+			return createInvalid(
+					"Issuer validation can not be performed because Jwt token does not contain a valid uri as 'jku' header parameter.");
+		}
+		if (!matchesTokenKeyUrlDomain(jkuUri)) {
+			return createInvalid(
+					"Issuer is not trusted because 'jku' '{}' does not match uaa domain '{}' of the identity service.",
+					jkuUri, domain);
+		}
+		if (!matchesTokenKeyEndpoint(jkuUri)) {
+			return createInvalid(
```
stefanvtSAP, line 58: only thing I could imagine is that a more detailed logging would be interessting. (e.g. because of URI parameters the jku does not match...)

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
```diff
@@ -38,27 +37,37 @@
...
-				tokenKeyUrl, domain);
+		return createValid();
+	}
+
+	private boolean matchesTokenKeyUrlDomain(URI jkuUri) {
+		return jkuUri.getHost() != null && jkuUri.getHost().endsWith(domain);
+	}
+
+	private boolean matchesTokenKeyEndpoint(URI jkuUri) {
+		return jkuUri.getPath().endsWith("token_keys") && jkuUri.getQuery() == null && jkuUri.getFragment() == null;
```
stefanvtSAP, line 70: I think you should better do:
`.endsWith("/token_keys") `
with the slash

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
nenaraab: createInvalid() logs the message with all details.

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtIssuerValidator.java
nenaraab: yeah, but path starts not with token_keys, so that is appropriate.

PR #314, opened 2020-06-24
subaccountid is provided with ext_attr instead of xs.system.attributes

nenaraab:


java-security/src/main/java/com/sap/cloud/security/token/XsuaaToken.java
```diff
@@ -159,6 +159,19 @@ public String getSubdomain() {
...
+	 * Returns subaccount identifier which is in most cases same like the
+	 * identity zone.
+	 * DO only use this for metering purposes.
+	 * DO NOT longer use this method to get the unique tenant id! For that use {@link #getZoneId()}.
+	 *
+	 * @return subaccount identifier or {@code null}
+	 */
+	@Nullable
+	public String getSubaccountId() {
+		return getAttributeFromClaimAsString(EXTERNAL_ATTRIBUTE, EXTERNAL_ATTRIBUTE_SUBACCOUNTID);
```
hassler-d, line 172: It is very similar to the implementation in `XSUserInfoAdapter` except it does not have a fallback to`ZONE_ID`: https://github.com/SAP/cloud-security-xsuaa-integration/pull/314/files#diff-47c999b39e9c3952559b7cade6014bd2R131 is this on purpose?
Can you define a test for this?

nenaraab: oh, yes - thanks for that review comment!

java-security/src/main/java/com/sap/cloud/security/token/XsuaaToken.java
```diff
@@ -159,6 +159,19 @@ public String getSubdomain() {
...
+	 * Returns subaccount identifier which is in most cases same like the
+	 * identity zone.
+	 * DO only use this for metering purposes.
+	 * DO NOT longer use this method to get the unique tenant id! For that use {@link #getZoneId()}.
+	 *
+	 * @return subaccount identifier or {@code null}
+	 */
+	@Nullable
+	public String getSubaccountId() {
+		return Optional.ofNullable(getAttributeFromClaimAsString(EXTERNAL_ATTRIBUTE, EXTERNAL_ATTRIBUTE_SUBACCOUNTID)).orElse(getZoneId());
```
hassler-d, line 172: @nenaraab I used `getZoneId()` as fallback. In `XSUserInfoAdapter` `getClaimValue(ZONE_ID)` is used. Which is not the same. Can we use getZoneId() here (and in `XSUserInfoAdapter`) as fallback?

nenaraab:
@hassler-d you need to perform another review :-)

java-security/src/main/java/com/sap/cloud/security/token/XsuaaToken.java
nenaraab: I would do it explizit as getZoneId() is not the right semantic... but ok for now.

hassler-d: No it is okay, it is just a small change. I just wanted to ask to get it consistent with `XSUserInfoAdapter`!

PR #320, opened 2020-06-30
Log token content/jwt validator config on failed authentication

hassler-d:
This is the simple solution for token logging more information on failed authentication attemps.

Todos:
- [x] AbstractToken.toString() provides token payload (maybe via DecodedJwt)
- [x] This should be logged in case validation fails.
- [x] Optional: log should provide validator that fails.
- [x] Optional: logging should be possible not only via Authenticator

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
+++ b/java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
@@ -190,4 +190,8 @@ public class DefaultJsonObject implements JsonObject {
 		}
 	}
 
+	@Override
+	public String toString() {
+		return jsonObject.toString(2);
+	}
 }
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtValidatorBuilder.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtValidatorBuilder.java
@@ -6,10 +6,13 @@ import com.sap.cloud.security.config.cf.CFConstants;
 import com.sap.cloud.security.token.Token;
 import com.sap.cloud.security.token.validation.CombiningValidator;
 import com.sap.cloud.security.token.validation.ValidationListener;
+import com.sap.cloud.security.token.validation.ValidationResult;
 import com.sap.cloud.security.token.validation.Validator;
 import com.sap.cloud.security.xsuaa.Assertions;
 import com.sap.cloud.security.xsuaa.client.*;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import java.util.*;
@@ -24,6 +27,8 @@ import static com.sap.cloud.security.config.cf.CFConstants.XSUAA.UAA_DOMAIN;
  * Custom validators can be added via {@link #with(Validator)} method.
  */
 public class JwtValidatorBuilder {
+	private static final Logger LOGGER = LoggerFactory.getLogger(JwtValidatorBuilder.class);
+
 	private static Map<OAuth2ServiceConfiguration, JwtValidatorBuilder> instances = new HashMap<>();
 	private final Collection<Validator<Token>> validators = new ArrayList<>();
 	private final List<ValidationListener> validationListeners = new ArrayList<>();
@@ -176,9 +181,25 @@ public class JwtValidatorBuilder {
 		List<Validator<Token>> allValidators = createDefaultValidators();
 		allValidators.addAll(validators);
 
-		CombiningValidator<Token> combiningValidator = new CombiningValidator<>(allValidators);
+		CombiningValidator<Token> combiningValidator = new CombiningTokenValidator(allValidators);
 		validationListeners.forEach(combiningValidator::registerValidationListener);
 		return combiningValidator;
+	}
+
+	private class CombiningTokenValidator extends CombiningValidator<Token> {
+
+		public CombiningTokenValidator(List<Validator<Token>> validators) {
+			super(validators);
+		}
+
+		@Override
+		public ValidationResult validate(Token token) {
+			ValidationResult validationResult = super.validate(token);
+			if (validationResult.isErroneous()) {
+				LOGGER.debug("Validation failed for token:{}{}", System.lineSeparator(), token);
+			}
+			return validationResult;
+		}
 	}
 
 	private List<Validator<Token>> createDefaultValidators() {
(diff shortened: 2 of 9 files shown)


hassler-d:
Now I have implemented the TODOs described above in a simple way without breaking changes.
This is how it looks like on failed validations now:
```log
[main] WARN com.sap.cloud.security.token.validation.ValidationResults - Jwt expired at 2019-10-26T03:32:49Z, time now: 2020-07-01T13:06:00.099Z
[main] DEBUG com.sap.cloud.security.token.validation.CombiningValidator - Validator that caused the failed validation: com.sap.cloud.security.token.validation.validators.JwtTimestampValidator
[main] DEBUG com.sap.cloud.security.token.validation.validators.JwtValidatorBuilder - Token that caused the failed validation: 
Jwt header
	{"alg":"RS256","jku":"https://authentication.stagingaws.hanavlab.ondemand.com/token_keys","kid":"key-id-1","typ":"JWT"}
Jwt payload
	{"jti":"2a6b47182ce34c83a3c4641b556b946c","ext_attr":{"enhancer":"XSUAA"},"sub":"sap_osb","authorities":["uaa.resource","ROLE_SERVICEBROKER"],"scope":["ROLE_SERVICEBROKER","uaa.resource"],"client_id":"sap_osb","cid":"sap_osb","azp":"sap_osb","grant_type":"client_credentials","rev_sig":"b23a72dc","iat":1572017569,"exp":1572060769,"iss":"http://localhost:8080/uaa/oauth/token","zid":"uaa","aud":["uaa","sap_osb"]}
```
I avoided adding the `Token` to the `ValidationResult` class because it would result in a lot of changes in the existing code. 
Furthermore the `ValidationResult` is used in the `Validator<T>` interface which is generic. If we add the `Token` to the  `ValidationResult` this would mean that every `ValidationResult` now contains a `Token` which only applies to validators that validate tokens. But since `Validator<T>` is generic, the ValidationResult that it uses  should not contain specifics. That would feel very strange. There are two solutions to this: (1) Make `ValidationResult` also generic over `T`, e.g. `ValidationResult<T>` (2) Get rid of the generics for the validation! There are only ever used for validating tokens anyway. 
By now I would prefer (2) but it would be a big breaking change. Option (1) could be implemented without a breaking change but I suspect the API would look better if this would be implemented with a breaking change also...
That is why I did not change the `ValidationResult`. I think we could change this someday (probably major release?). But it should not be part of this PR!

java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
```diff
@@ -190,4 +190,8 @@ private JSONObject createJsonObject(String jsonString) {
 		}
 	}
 
+	@Override
+	public String toString() {
+		return jsonObject.toString(2);
```
nenaraab, line 195: what does toString(2)?

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtValidatorBuilder.java
```diff
@@ -176,11 +181,27 @@ public JwtValidatorBuilder withValidatorListener(ValidationListener validationLi
 		List<Validator<Token>> allValidators = createDefaultValidators();
 		allValidators.addAll(validators);
 
-		CombiningValidator<Token> combiningValidator = new CombiningValidator<>(allValidators);
+		CombiningValidator<Token> combiningValidator = new CombiningTokenValidator(allValidators);
 		validationListeners.forEach(combiningValidator::registerValidationListener);
 		return combiningValidator;
 	}
 
+	private class CombiningTokenValidator extends CombiningValidator<Token> {
```
nenaraab, line 189: why is this private class required? there is already a public  CombiningValidator?

java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
hassler-d: It pretty prints the json object with indentation level 2

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtValidatorBuilder.java
hassler-d: Moved logging into CombiningValidator<Token>

hassler-d:
@nenaraab I tried to update the approuter to 8.1.1 but it failed with
```log
Updating app approuter-sap-java-buildpack-api-usage...
Mapping routes...
Comparing local files to remote cache...
Packaging files to upload...
Uploading files...
 5.43 KiB / 5.43 KiB [==========================================================================] 100.00% 1s

Waiting for API to complete processing files...

Stopping app...

Staging app and tracing logs...
   Downloading nodejs_buildpack...
   Downloaded nodejs_buildpack
   Cell 51b5c234-d374-42ca-9800-8415875163d6 creating container for instance 57030796-dc09-4607-9684-f25edc9931e3
   Cell 51b5c234-d374-42ca-9800-8415875163d6 successfully created container for instance 57030796-dc09-4607-9684-f25edc9931e3
   Downloading build artifacts cache...
   Downloading app package...
   Downloaded app package (5.4K)
   Downloaded build artifacts cache (11.8M)
   -----> Nodejs Buildpack version 1.7.15
   -----> Installing binaries
          engines.node (package.json): unspecified
          engines.npm (package.json): unspecified (use default)
          **WARNING** Node version not specified in package.json or .nvmrc. See: http://docs.cloudfoundry.org/buildpacks/node/node-tips.html
   -----> Installing node 10.19.0
          Copy [/tmp/buildpacks/eefd3f0701470f5b366a9eaaec589c68/dependencies/6490223e294e93b931ffbc8141f8d8b1/node-10.19.0-linux-x64-cflinuxfs3-0bcac686.tgz]
          Using default npm version: 6.13.4
   -----> Installing yarn 1.22.4
          Copy [/tmp/buildpacks/eefd3f0701470f5b366a9eaaec589c68/dependencies/d9f0c5147bcb9ee798f041978c157822/yarn-1.22.4-any-stack-24ca2294.tgz]
          Installed yarn 1.22.4
   -----> Creating runtime environment
          PRO TIP: It is recommended to vendor the application's Node.js dependencies
          Visit http://docs.cloudfoundry.org/buildpacks/node/index.html#vendoring
          NODE_ENV=production
          NODE_HOME=/tmp/contents924408020/deps/0/node
          NODE_MODULES_CACHE=true
          NODE_VERBOSE=false
          NPM_CONFIG_LOGLEVEL=error
          NPM_CONFIG_PRODUCTION=true
   -----> Building dependencies
          Installing node modules (package.json)
Error staging application: App staging failed in the buildpack compile phase
FAILED
```

Do you know which version we can use for the sample?

nenaraab:
hi @hassler-d 
you can find the latest one here:
https://www.npmjs.com/package/@sap/approuter

PR #324, opened 2020-07-03
Fix wrong DEFAULT_JWKS_URL in JwtGenerator

hassler-d:
This also changes the existing validation test so that the defaults of the JwtGenerator are not overridden in the test.


java-security-test/src/test/resources/vcap.json
```diff
@@ -8,8 +8,8 @@
      "identityzone": "uaa",
      "identityzoneid": "uaa",
      "tenantmode": "dedicated",
-     "uaadomain": "auth.com",
-     "url": "https://paastenant.auth.com",
+     "uaadomain": "localhost",
```
nenaraab, line 11: Hi Daniel,

i would remain this, because this checks other parts of the validation. Can't you just add another test-vcap-services.

hassler-d: I have reversed those changes, added a new test and moved the integration tests into a new test class.

PR #325, opened 2020-07-06
XsuaaToken should provide method to access external attributes

nenaraab:


nenaraab:
@hassler-d 
good point, i will provide that as part of the API as it is more generic....
getSubdomain/getSubaccountId remains in XsuaaToken w/o API

PR #333, opened 2020-07-14
AccessToken getAttributeFromClaimAsStringList

nenaraab:
... to access xs.user.attributes


nenaraab:
provide another one without python script changes

PR #335, opened 2020-07-14
Latest version available on central is 2.7.4

raulnegrean:


nenaraab:
we had some technical issue during release. Version 2.7.5 is now available on maven central :-)

PR #336, opened 2020-07-14
Python script 2fa

nenaraab:
export ENABLE_2_FACTOR=x


hassler-d:
> Works on my machine as well. All tests are green!

I have tested locally with #337, #336 and #334 merged.

PR #342, opened 2020-07-17
Cas update

nenaraab:
- simplify Auto-configuration MethodSecurityConfiguration in that regard that the interface with lazy load is not required anymore
- rename Autoconfiguation classes
  - AdcServiceConfiguration --> AdcServiceAutoConfiguration
  - MethodSecurityConfiguration 	--> AdcSpringMethodSecurityAutoConfiguration
- AdcSpringSecurityExpression should not have any dependency to spring-security-oauth2-resource-server
  - dependency should only be provided as part of pom.xml

@hassler-d can you please test this with your xsuaa sample?

### DIFF
--- a/spring-cas/pom.xml
+++ b/spring-cas/pom.xml
@@ -47,6 +47,7 @@
 		<dependency>
 			<groupId>org.springframework.security</groupId>
 			<artifactId>spring-security-oauth2-resource-server</artifactId>
+			<scope>provided</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.security</groupId>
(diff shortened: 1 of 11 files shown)


spring-cas/pom.xml
```diff
@@ -47,6 +47,7 @@
 		<dependency>
 			<groupId>org.springframework.security</groupId>
 			<artifactId>spring-security-oauth2-resource-server</artifactId>
+			<scope>provided</scope>
```
hassler-d, line 50: I tested this with both samples (spring-security-cas and the sample with xsuaa) and it is working!

spring-cas/src/main/java/com/sap/cloud/security/cas/spring/AdcSpringSecurityExpression.java
```diff
@@ -29,19 +26,11 @@
 	private String userId;
```
hassler-d, line 26: can be moved inside following block

spring-cas/src/main/java/com/sap/cloud/security/cas/spring/AdcSpringSecurityExpressionHandlerXsuaa.java
```diff
@@ -23,9 +34,16 @@ public static AdcSpringSecurityExpressionHandlerXsuaa getInstance(AdcService ser
 	@Override
 	protected MethodSecurityExpressionOperations createSecurityExpressionRoot(
 			Authentication authentication, MethodInvocation invocation) {
+		String zoneId;
```
hassler-d, line 37: can be moved inside following block

PR #357, opened 2020-08-24
Integration test utilities

hassler-d:
This PR adds integration test utilities to improve integration testing.
It enhances the existing JwtGenerator and adds VcapServicesParser class.


java-security-it/src/test/java/com/sap/cloud/security/test/integration/XsuaaIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+		ValidationResult result = tokenValidator.validate(token);
+		assertThat(result.isValid()).isTrue();
+	}
+
+	@Test
+	public void xsaTokenValidationSucceeds_withXsuaaCombiningValidator() throws IOException {
+		OAuth2ServiceConfiguration configuration = VcapServicesParser.fromFile(
+				"/xsuaa/vcapServices/xsaSingleBinding.json")
+				.runInLegacyMode(true)
+				.withUrl(rule.getWireMockServer().baseUrl()) // overridden because of legacy mode
```
nenaraab, line 43: I do not understand why this is needed? overridden because of legacy mode?

java-security-it/src/test/resources/xsuaa/vcapServices/xsaSingleBinding.json
```diff
@@ -0,0 +1,18 @@
+{
+  "xsuaa" : [ {
+    "name" : "java-uaa",
+    "label" : "xsuaa",
+    "tags" : [ "xsuaa" ],
+    "plan" : "space",
+    "credentials" : {
+      "tenantmode" : "dedicated",
+      "clientid" : "sb-java-hello-world!i1",
+      "verificationkey" : "-----BEGIN PUBLIC KEY-----MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwOzNGEDK6jNrJcl3L2wstY8SnX9KNo4ic/JtrEQX4U/+749b9y6JlSGOXSKBFJ6alNGHVyPK66SfSgeWHwPjVekdnHYM7bQTyYX1NoSOxqUgvsktIpxeYJLC+J5qJTtnFcjFTRUovKwtCYxlTP4b63b7dWixwc+iSnkuvl4IkZjVTcstK73eTlmVtJDC5DIisPyYOK06qxTPrLC8KlgFA6a0FdPJF9G8s/ElxPHFu7mw6UKn2LTmDlDbADQZLMXGKBw/zg3l5scjxkOCVGTkGvv1ooccqCfYgZAsmsPbQdoZxznSdQTPfaWULnYLo8i5hFXiS1q8agvXHjWpUmWcMwIDAQAB-----END PUBLIC KEY-----",
```
nenaraab, line 10: we can remove that... as it is anyway ignored

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -158,23 +153,60 @@ public JwtGenerator withClaimValues(String claimName, String... values) {
...
+	 * </pre>
+	 *
+	 * The payload and header data from the file will be written into the token that
+	 * is being generated. Note that some properties from the file are ignored. This
+	 * is for convenience so that the token can be verified in a test setup rather
+	 * then its original production setup. The following header and payload
+	 * properties are ignored:
+	 * <ul>
+	 * <li>Header: jku, kid</li>
+	 * <li>Payload: exp, aud, iss</li>
```
nenaraab, line 186: aud is not longer ignored... only in case it was not filled, it is set....

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -158,23 +153,60 @@ public JwtGenerator withClaimValues(String claimName, String... values) {
...
+	 *             if the file does not contain a valid json object
+	 * @throws IOException
+	 *             when the file cannot be read or does not exist.
+	 */
+	public JwtGenerator fromFile(String tokenJsonResource) throws IOException {
+		String tokenJson = IOUtils.resourceToString(tokenJsonResource, StandardCharsets.UTF_8);
+		JSONObject jsonObject = createJsonObject(tokenJson);
+		copyJsonProperties(filterPayload(jsonObject.optJSONObject("payload")), jsonPayload);
+		copyJsonProperties(filterHeader(jsonObject.optJSONObject("header")), jsonHeader);
+		if (GrantType.CLIENT_CREDENTIALS == getGrantType()) {
```
nenaraab, line 206: we should discuss that clientId again... related claims: sub, aud, cid, client_id

nenaraab (changes requested):
Hi @hassler-d 
thanks a lot for this nice PR!

These are my general statements....

1. I think we have to rethink again the spec, as I feel it is still too complicated / confusing...
  - in context of integration tests i think it is not so relevant to make sure, that the token-claims fits to the service config. So i guess its more relevant to log warnings, if they do not fit together...
  - so, when loading a token from file, these fields should NOT be overwritten: `cid` / `client_id`, `sub` and audience.
  - BUT, when using mockserver / securitytestrule the urls have to be overwritten within the service config and the token, this still makes sense.
2. Maybe it then makes sense, if we reorganise the resource files once more:
  - src/test/resources/ias-simple/vcap_services-single.json and .../token.txt
  - src/test/resources/xsa-simple/vcap_services.json and .../token.txt

Thanks a lot!
Nena

java-security-it/src/test/java/com/sap/cloud/security/test/integration/IasIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IasIntegrationTest {
+
+	@ClassRule
+	public static SecurityTestRule rule = SecurityTestRule.getInstance(Service.IAS)
+			.setKeys("/publicKey.txt", "/privateKey.txt");
```
nenaraab, line 22: which token keys do yu make use of here? as IAS token keys looks slightly different...

Furthermore is there a central place where these mock server endpoints are configured?
See also here:
https://github.wdf.sap.corp/CPSecurity/Knowledge-Base/blob/master/99_Internal/03_ApplicationSecurity/IntegrationTests/README.md#mock-server-oidc-discovery-endpoint

nenaraab (changes requested):

pom.xml
```diff
@@ -86,6 +86,7 @@
 		<module>java-api</module>
 		<module>java-security</module>
 		<module>java-security-test</module>
+		<module>java-security-it</module>
```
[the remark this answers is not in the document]
hassler-d: Got it. Makes sense!

java-security-test/src/test/resources/client_token.json
```diff
@@ -0,0 +1,8 @@
+{
+  "header": {
```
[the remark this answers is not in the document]
hassler-d: Empty header is OK. If it is left out we have not defined what happens. Right now it would be an NullPointer which is bad. What should happen? Also just ignore it?

java-security-it/src/test/java/com/sap/cloud/security/test/integration/IasIntegrationTest.java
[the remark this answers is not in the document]
hassler-d: No. By default generated keys would be used which also works.

java-security-it/src/test/java/com/sap/cloud/security/test/integration/XsuaaIntegrationTest.java
hassler-d: In this test the validation runs in legacy mode. In legacy mode the URL for the token_keys endpoint is read from the URL parameter of the configuration instead of the JKU header parameter. That is why the url needs to be overridden with the correct `baseUrl` of the wire mock server.

java-security-it/src/test/resources/xsuaa/vcapServices/xsaSingleBinding.json
hassler-d: True, removed!

java-security-test/src/test/resources/client_token.json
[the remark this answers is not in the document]
nenaraab: Yes, no header should behave the same like "empty" one to make it a little bit more robust.

java-security-it/src/test/java/com/sap/cloud/security/test/integration/IasIntegrationTest.java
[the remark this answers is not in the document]
nenaraab: Then, I would remove it... but preferably with a IAS specific key... they looks a little bit different than the xsuaa keys... isn't it?

java-security-it/src/test/java/com/sap/cloud/security/test/integration/XsuaaIntegrationTest.java
nenaraab: Is the Mockserver setup done in SecurityTestRule? And is there the place, where the urls... (iss, config.url) are replaced?
If yes, we could think about moving the legacyMode flag into that... probably we need then another IT-class...?

hassler-d: The Mockserver setup is done in SecurityTestRule. But the SecurityTestRule does not know anything about the service configuration or the VcapServicesParser. Therefore this url needs to be replaced inside the test where all the information is available. The challenge here is that the wireMockServer URL is only available after it has been started.

I also think that it would be easier if we had a component that combines the SecurityTestRule and the configuration object and makes sure that the settings of those two match. Another possibility would be to integrate the configuration and the configuration loading mechanism into the SecurityTestRule. But this would increase the size of the SecurityTestRule further and it is already pretty big.

java-security-it/src/test/java/com/sap/cloud/security/test/integration/IasIntegrationTest.java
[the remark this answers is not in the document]
hassler-d: IAS tokens are signed differently than XSUAA tokens?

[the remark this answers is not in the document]
nenaraab: No. But the token_keys endpoint provides the token differently. This can be managed with different key-ids....
XSUAA:
https://green-1.authentication.sap.hana.ondemand.com/token_keys
IAS:
https://xs2security.accounts400.ondemand.com/oauth2/certs

So, finally i would like to have both of them in preferably one file, e.g. token_keys.txt with two different entries one as default for ias and another as default for xsuaa...

java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTest.java
```diff
@@ -183,26 +186,68 @@ public SecurityTest setKeys(String publicKeyPath, String privateKeyPath) {
...
 					.withHeaderParameter(TokenHeader.JWKS_URL, jwksUrl)
 					.withAppId(DEFAULT_APP_ID)
 					.withClaimValue(TokenClaims.XSUAA.GRANT_TYPE, OAuth2TokenServiceConstants.GRANT_TYPE_USER_TOKEN);
-			break;
-		default:
-			jwtGenerator.withClaimValue(TokenClaims.ISSUER, wireMockServer.baseUrl());
-			break;
 		}
-		return jwtGenerator;
+		return jwtGenerator.withClaimValue(TokenClaims.ISSUER, issuerUrl);
```
nenaraab, line 206: this seems to be an incompatible change, isn't it?

nenaraab (approved):

PR #358, opened 2020-08-25
Json object.get as string list

nenaraab:


java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
```diff
@@ -47,6 +47,19 @@ public boolean isEmpty() {
...
 	}
 
+	@Override
+	public List<String> getAsStringList(String name) {
+		List<String> list = new ArrayList<>();
+		if(contains(name)) {
+			if (getJsonObject().get(name) instanceof String) {
+				list.add(getAsString(name));
+			} else {
+				list = getJSONArray(name).map(jsonArray -> castToListOfType(jsonArray, String.class)).orElse(Collections.emptyList());
```
hassler-d, line 57: You could use getAsList in this line, e.g.  list = getAsList(name, String.class)

hassler-d (approved):
Looks good to me. I think one line can be simplified, see my comment!

java-security/src/main/java/com/sap/cloud/security/json/DefaultJsonObject.java
nenaraab: yes, you're right... i forgot that after i've moved it from getAsList method into a the new getAsStringList() :-)

Issue #359, opened 2020-08-25
Feature request: Configuration option whether to accept or reject broker clone tokens


thomas-br:
## Current State
In the current state of the [XsuaaJwtDecoder](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java) and the [XsuaaAudienceValidator](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidator.java), access tokens issued from binding information of clones of xsuaa instances with the broker-plan (here called as `broker clone tokens` or `clone tokens`) are accepted by default. With the broker plan clones can be created from a master xsuaa instance, that are bound to a specific org / tenant and enable tenant specific technical authentication. As an example the binding information / credentials could be given to the respective tenants to enable API access instead of a pure browser flow access based on the IDP configuration.

## Gap Scenario
The current state implies that these clone tokens are automatically accepted as fully trusted and authenticated tokens which holds true from pure broker scenarios. In other application setup scenarios a mixed form is also valid, that API access exists besides browser scenarios. But those two scenarios do have a different trust levels, as the entry-point for browser access and user tokens is always the approuter. Having existing clone binding credentials provided to tenants moves the token control to the outside. 

In specific setups it is required to reject `clone-tokens` per service, which might not be intended to be called directly but rather only with tokens coming from the "xsuaa master instance / binding".

## Requested Feature
Offer a configuration parameter, for example through the [XsuaaJwtDecoderBuilder](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoderBuilder.java) to decide whether these `clone-tokens` should be accepted (default case) or rejected instead on a service level.

nenaraab:
Hi @torsten-sap can you please evaluate this use case... and if this is valid... how shall we call the feature?

nenaraab:
Hi @thomas-br 

I've discussed internally and another idea came up to reflect _the different trust levels_:
- the master could set "authorities-inheritance": false in order to not inherit the master scopes to the clones.

In that case, the AudienceValidation will not fail, but the scope check would fail. Would that be feasible?

[...]

PR #361, opened 2020-08-25
Feature request: Configuration option whether to accept or reject broker clone tokens

nenaraab:
#359


nenaraab:
this solution doesn't seem to be the right one. It gets replaced with PR #369

Issue #340, opened 2020-07-16
Feature request: inspect JWT caching effectiveness


patrickmhaller:
Hi team,

thank you very much for introducing JWT caching with https://github.com/SAP/cloud-security-xsuaa-integration/blob/4fd0366bdc1da01b8df9f6a46db18d12b84537d2/token-client/src/main/java/com/sap/cloud/security/xsuaa/tokenflows/CacheConfiguration.java#L14 !

Could you kindly add a way to access these internals
* configure the cache to record statistics
* export the statistics ( calls, hits, misses, refreshes, avg. response time from XSUAA) so they can be observed by a caller or even exported to Dynatrace via JMX

I currently could not see a way to influence `AbstractOAuth2TokenService.createResponseCache` in that way.

Thanks,
Patrick

nenaraab:
Hi @patrickmhaller 

I've understood, you like to have that for dynatrace monitoring. 
What is exactly would your requirement in regard to export the statistics.

Best regards,
Nena

patrickmhaller:
Hi @nenaraab,

First, we'd like to be able to optionally enable statistics collection:
cache = Caffeine.newBuilder()
&rarr;        **.recordStats()**        &larr;
        .build());

Secondly, we'd need access to the statistics and the native cache object, hence like
```
        final com.github.benmanes.caffeine.cache.Cache nativeCache = (com.github.benmanes.caffeine.cache.Cache) cache;
        final CacheStats                               stats       = nativeCache.stats();
```
nativeCache.estimatedSize() can tell about the size of the cache at a given time,
stats holds all the effectiveness and efficiency statistics we're looking for, even totalLoadTime, averageLoadPenalty, 

With this, we can then expose these statistics via a JMX bean to Dynatrace in our coding.

This pretty much explains the idea, I hope.
We have found that XSUAA token requests usually take between 40ms and 200ms, averaging at ~120ms in production. The fact that we have very often to call secondary services needing a token switch is much amplifying the impact on overall responsiveness. Therefore, effective JWT caching is crucial for us, hence the desire for monitoring.

Thanks,
Patrick

hassler-d:
Hi Patrick,

we have implemented this in #367 and it will be part of the upcoming 2.7.7. release.

Best Regards
Daniel

PR #367, opened 2020-09-07
Cache statistics

hassler-d:
This PR adds the possibility to enable cache statistics in the `CacheConfiguration`. Every `Cacheable` can return their specific cache statistic object. 
This PR is motivated by #340


nenaraab (approved):
@hassler-d looks as defined. W/o incompatible changes... Thanks a lot!

PR #374, opened 2020-09-24
Set key id depending on service

hassler-d:
Respect service when setting key id on token key endpoint.
Added static key and changed key id of static keys.


nenaraab (approved):
LGTM :-)

PR #385, opened 2020-10-06
Simplifications of XsuaaAudienceValidator

eaglerainbow:
Recently, I had to debug through `XsuaaAudienceValidator` and I felt that it was partially hard to read. In method `getAllowedAudiences` I had the impression that you have a hard time handling things using Java Lists, but in fact do not want to use the property of ordering at all. Instead, you are busy fixing things afterwards using Java Streams.

Therefore, I'd like to propose this PR making things simpler to read without losing any kind of functionality or quality. The general idea is to not make use of Lists, but to use a (Hash)Set instead. This also sightly will have advantages in terms of performance (in case JWTs do get larger).

Feel free to merge; I do not claim any ownership of my changes, but only want to make things better...

### DIFF
--- a/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidatorTest.java
+++ b/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidatorTest.java
@@ -7,6 +7,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Set;
 
 import org.junit.Assert;
 import org.junit.Before;
@@ -67,8 +68,7 @@ public class XsuaaAudienceValidatorTest {
 	public void testExtractAudiencesFromTokenScopes() {
 		Jwt token = new JwtGenerator()
 				.addScopes("test1!t1.read", "test2!t1.read", "test2!t1.write", ".scopeWithoutAppId").getToken();
-		List<String> audiences = new XsuaaAudienceValidator(serviceConfigurationSameClientId)
-				.getAllowedAudiences(token);
+		Set<String> audiences = XsuaaAudienceValidator.getAllowedAudiences(token);
 		Assert.assertThat(audiences.size(), is(2));
 		Assert.assertThat(audiences, hasItem("test1!t1"));
 		Assert.assertThat(audiences, hasItem("test2!t1"));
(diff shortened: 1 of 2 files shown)


eaglerainbow:
@nenaraab  might want to have a look at it.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidatorTest.java
```diff
@@ -67,8 +68,7 @@ public void testSameClientIdWithoutAudience() {
 	public void testExtractAudiencesFromTokenScopes() {
 		Jwt token = new JwtGenerator()
 				.addScopes("test1!t1.read", "test2!t1.read", "test2!t1.write", ".scopeWithoutAppId").getToken();
-		List<String> audiences = new XsuaaAudienceValidator(serviceConfigurationSameClientId)
-				.getAllowedAudiences(token);
+		Set<String> audiences = XsuaaAudienceValidator.getAllowedAudiences(token);
```
nenaraab, line 71: This is not required. Please keep existing tests as it is.

nenaraab (commented):
Hi @eaglerainbow 

thanks a lot for the code improvements. I like them.
Please undo the change in the test code, and run "mvn formatter:format".

Thanks again and best regards,
Nena

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidatorTest.java
eaglerainbow: Well, I don't agree to that end:
1. I must adjust the return type of the method in any case.
2. The method under test is declared as `static` - and hence should be accessed not through an instance, but better through the class.

I would be willing to revert item 2 (though I still think it's wrong), but reverting item 1 would break your test.

Please comment on what you want to have :)

eaglerainbow:
> and run "mvn formatter:format"

done with 7190d0de

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidatorTest.java
nenaraab: Hi @eaglerainbow,

You‘re right, thanks for the hint! I will rename the test afterwards...

Best regards,
Nena

nenaraab (approved):

PR #387, opened 2020-10-07
Add test that checks stubbing works after setup

hassler-d:
This added test case checks that it is possible to stub an endpoint using the `SecurityTestContext` after the underlying wiremock server has been already been started.


liga-oz (approved):
All good 👍

nenaraab (approved):
LGTM

PR #389, opened 2020-10-08
Change request: Replace "cid" with "azp" / "aud"

liga-oz:
Instead of `cid` (OAuth2 standard) use OIDC standard, that means
- return `azp` claim if claim exists with non-null value 
- return `aud` claim (which is single String or StringList with only one entry)
- return error in all the other cases

Spec https://openid.net/specs/openid-connect-core-1_0.html


java-api/src/main/java/com/sap/cloud/security/token/TokenClaims.java
```diff
@@ -30,7 +30,7 @@ private XSUAA() {
 		public static final String ORIGIN = "origin";
 		public static final String GRANT_TYPE = "grant_type"; // OAuth grant type used for token creation
 		public static final String ZONE_ID = "zid"; // tenant GUID -> SAP_GLOBAL_ZONE_ID
-		public static final String CLIENT_ID = "cid"; // OAuth client identifier
+		public static final String CLIENT_ID = "azp"; // OAuth client identifier
```
nenaraab, line 33: @liga-oz you can keep that and set it to `@deprecated` with a javadoc hint `@link #azp` field

and add a new one for azp

java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -149,4 +149,11 @@
 	 */
 	String getZoneId();
 
+	/**
+	 * Returns Client identifier,
```
nenaraab, line 153: document the oidc standard - azp -> aud....

java-security/src/main/java/com/sap/cloud/security/token/ClientIdRetrievalException.java
```diff
@@ -0,0 +1,10 @@
+package com.sap.cloud.security.token;
+
+/**
+ * Runtime exception during client id retrieval from token.
+ */
+public class ClientIdRetrievalException extends RuntimeException {
```
nenaraab, line 6: -> TokenMisconfigurationException --> just move to java-api module

nenaraab (commented):
only minor comments from my side

hassler-d:
Hi @nenaraab can you take a look at it again? We went all in on `azp` in the `JwtGenerator(s)` and we weren't sure if this is the right approach or if the `azp` parameter should just be used internally.

liga-oz:
deploy_and_test script.py execution was successful

spring-xsuaa-test/src/test/java/com/sap/cloud/security/xsuaa/test/JwtGeneratorTest.java
```diff
@@ -19,13 +19,13 @@
 
 public class JwtGeneratorTest {
 	private JwtGenerator jwtGenerator;
-	private static final String MY_CLIENT_ID = "sb-clientId!20";
+	private static final String AUTHORIZATION_PARTY = "sb-clientId!20";
```
nenaraab, line 22: AUTHORIZATION_PARTY -> MY_CLIENT_ID

java-api/src/main/java/com/sap/cloud/security/token/TokenClaims.java
```diff
@@ -30,7 +30,12 @@ private XSUAA() {
 		public static final String ORIGIN = "origin";
 		public static final String GRANT_TYPE = "grant_type"; // OAuth grant type used for token creation
 		public static final String ZONE_ID = "zid"; // tenant GUID -> SAP_GLOBAL_ZONE_ID
+		/**
+		 * @deprecated Use {@link Token#getClientId} instead.
+		 */
+		@Deprecated
 		public static final String CLIENT_ID = "cid"; // OAuth client identifier
+		public static final String AUTHORIZATION_PARTY = "azp"; // Authorization party contains OAuth client identifier
```
nenaraab, line 38: pls move to generic part...

spring-xsuaa-test/src/test/java/com/sap/cloud/security/xsuaa/test/JwtGeneratorTest.java
nenaraab, line 22: AUTHORIZATION_PARTY -> MY_CLIENT_ID

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/XsuaaToken.java
```diff
@@ -141,7 +143,19 @@ public String getLogonName() {
...
 	public String getClientId() {
-		return getClaimAsString(CLAIM_CLIENT_ID);
+		String clientId = getClaimAsString(AUTHORIZATION_PARTY);
+		if (clientId == null || clientId.trim().isEmpty()) {
+			List<String> audiences = getAudience();
+
+			if (audiences != null && audiences.size() == 1) {
+				return audiences.get(0);
+			} else {
+				logger.error("Couldn't get client id. Invalid authorized party or audience claims.");
```
nenaraab, line 153: check for cid (and add a comment that this is only required for support generated tokens in JUnit tests. Finally they should provide "azp" claim instead of "cid" claim) and then als log a WARNING

java-security/src/main/java/com/sap/cloud/security/token/AbstractToken.java
```diff
@@ -162,7 +163,24 @@ public int hashCode() {
...
+	@Override
+	public String getClientId() {
+		String clientId = getClaimAsString(AUTHORIZATION_PARTY);
+		if (clientId == null || clientId.trim().isEmpty()) {
+			Set<String> audiences = getAudiences();
+
+			if (audiences.size() == 1) {
+				return audiences.stream().findFirst().get();
+			} else {
+				LOGGER.error("Couldn't get client id. Invalid authorized party or audience claims.");
```
nenaraab, line 178: check for cid (and add a comment that this is only required for support generated tokens in JUnit tests. Finally they should provide "azp" claim instead of "cid" claim) and then als log a WARNING

nenaraab (changes requested):
add the fallabck for cid and some minor comments.

nenaraab (changes requested):
Hi @liga-oz 

i've reviewed again and "reverted" some of the former test cases (especially in JwtGenerator) to guarantee the compatibility we have agreed on.

One incompatible i see, we have to check after my vacation and we have to document is:
when people use JwtGenerator (java-security-test) and like to "overwrite" the XSUAA.CLIENT_ID => this has to be changed by them to XSUAA.AUTHORIZATION_PARTY

Furthermore I recognized, that XSUserInfoAdapter.getClientId() is not yet changed. But it needs to be adpated as well to:
```
accessToken.getClientId();
```
With that, I guess its done :-)

Thanks a lot!
Nena

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -172,7 +172,9 @@ public JwtGenerator withHeaderParameter(String parameterName, String value) {
 	}
 
 	/**
-	 * Sets the claim with the given name to the given string value.
+	 * Sets the claim with the given name to the given string value. Note: for
```
nenaraab, line 175: as this is the withHeaderParameter method, this javadoc comment can be removed.

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -186,7 +188,9 @@ public JwtGenerator withClaimValue(String claimName, String value) {
...
```
nenaraab, line 191: +1

nenaraab (changes requested):

java-security/src/test/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapterTest.java
```diff
@@ -379,7 +379,7 @@ public void getHdbToken_AudCodeToken_WithAttributes() throws XSUserInfoException
 		String mockTokenValue = "mock token value";
 
 		when(token.getTokenValue()).thenReturn(mockTokenValue);
-		when(token.getClaimAsString(TokenClaims.XSUAA.CLIENT_ID)).thenReturn("sb-margin-assurance-ui!i1");
+		when(token.getClientId()).thenReturn("sb-margin-assurance-ui!i1");
```
nenaraab, line 382: pls undo changes on test. They should run compatible.

nenaraab (changes requested):

CHANGELOG.md
```diff
@@ -13,7 +12,11 @@ All notable changes to this project will be documented in this file.
   - javax.servlet-api.version 3.0.1 --> 3.1.0
   - Apache HTTP client 4.5.8 --> 4.5.9
   - Wiremock 2.25.1 --> 2.27.2
-  
+- [java-security]
```
nenaraab, line 15: yeah, this can be moved to a new section "2.8.0"

java-security-test/src/test/java/com/sap/cloud/security/test/JwtGeneratorTest.java
```diff
@@ -335,7 +335,7 @@ public void getInstanceFromFile_loadsJsonData() throws IOException {
 		assertThat(token.getClaimAsString(TokenClaims.XSUAA.ZONE_ID)).isEqualTo("zone-id");
 		assertThat(token.getClaimAsStringList(TokenClaims.XSUAA.SCOPES)).containsExactlyInAnyOrder("openid",
 				"app1.scope");
-		assertThat(token.getClaimAsString(TokenClaims.XSUAA.CLIENT_ID)).isEqualTo("testingClientId");
+		assertThat(token.getClientId()).isEqualTo("testingClientId");
```
nenaraab, line 338: actually you can test both ?
add this test case as well?

`assertThat(token.getClaimAsString(TokenClaims.XSUAA.CLIENT_ID)).isEqualTo("testingC`lientId");

nenaraab (changes requested):

java-security/src/test/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapterTest.java
liga-oz: @nenaraab to revert this back, I would need to also mock hasClaim(CLIENT_ID) and call real method of getClientId() to comply with the new getClientId impl. Would that suit better?

nenaraab (approved):

PR #392, opened 2020-10-14
Add junit-bom to dependency management

hassler-d:
Adding the junit-platform-commons to the dependency management ensures that this specified version will be used. Without explicitly setting the version, third party dependencies might bring junit-platform-commons in an older version. This causes problem when this older version is not compatible with junit-jupiter used in this project.

With this change the performance tests can be run again from java-security-it via `mvn failsafe:integration-test`

**EDIT: I now use the junit 5 bom, see newest [comment](https://github.com/SAP/cloud-security-xsuaa-integration/pull/392#issuecomment-708341269)**

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,8 +72,6 @@
 		<javax.servlet.api.version>3.1.0</javax.servlet.api.version>
 		<mockwebserver.version>3.9.1</mockwebserver.version>
 		<junit.version>4.13.1</junit.version>
-		<!-- make sure junit-platform and junit-jupiter versions are compatible -->
-		<junit-platform.version>1.7.0</junit-platform.version>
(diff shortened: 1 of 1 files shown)


hassler-d:
I could not get the integration tests to run by putting the jupiter dependencies to the front of the pom.xml. It still pulled the wrong dependency version. 
However, I have found the junit 5 [bill of materials](https://mvnrepository.com/artifact/org.junit/junit-bom/5.7.0) which I replaced all junit 5 dependencies in the main pom.xml with.
This also solves the problem and now there is only one dependency for junit5 in the main pom.xml. 
@nenaraab would this solution work out for you?

nenaraab (approved):

PR #412, opened 2020-11-06
JwtGenerator should prefill the zone and user with defaults

nenaraab:
- IAS provides "zone_uuid" and "user_uuid" 
- XSUAA provides as of now only "zid"

These fields should be prefilled. Applications can override if needed.


liga-oz (approved):
Looking good

PR #421, opened 2020-11-27
Update README.md

liga-oz:


nenaraab (commented):
Hi Liga, like the idea, but then i would recommend to adapt all places and check whether no link is broken, as the implicit link to #known-issues will be deleted.

nenaraab (approved):

PR #431, opened 2020-12-11
Ias xsuaa token exchange

liga-oz:


java-security/src/main/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticator.java
```diff
@@ -39,9 +64,83 @@ protected TokenAuthenticationResult authenticated(Token token) {
...
 
+    @Override
+    public TokenAuthenticationResult validateRequest(ServletRequest request, ServletResponse response) {
+        if (request instanceof HttpServletRequest && response instanceof HttpServletResponse) {
+            HttpServletRequest httpRequest = (HttpServletRequest) request;
+            String authorizationHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
+            if (headerIsAvailable(authorizationHeader)) {
+                try {
+                    Token token = extractFromHeader(authorizationHeader);
+                    if (isIasXsuaaXchangeEnabled() && !isXsuaaToken(token)) {
```
nenaraab, line 75: you can even put the `isIasXsuaaXchangeEnabled()` check to the first if-clause.

java-security/src/main/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticator.java
```diff
@@ -39,9 +64,83 @@ protected TokenAuthenticationResult authenticated(Token token) {
...
+     * @return true if provided token is a XSUAA token
+     */
+    private boolean isXsuaaToken(Token token) {
+        try {
+            JsonObject externalAttributeClaim = token.getClaimAsJsonObject(EXTERNAL_ATTRIBUTE);
+            String externalAttributeValue = Objects.requireNonNull(externalAttributeClaim, "ext_attr claim is missing")
+                    .getAsString(EXTERNAL_ATTRIBUTE_ENHANCER);
+            logger.debug("Token attributes claims: {}: {}", EXTERNAL_ATTRIBUTE, externalAttributeClaim);
+            return Objects.requireNonNull(externalAttributeValue, "enhancer is missing").equalsIgnoreCase("xsuaa");
+        } catch (NullPointerException e) {
```
nenaraab, line 120: I would not catch NullPointerException here, instead to a nullpointer check.

java-security/src/test/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticatorTest.java
```diff
@@ -138,6 +148,67 @@ public void validateRequest_invalidToken_listenerIsCalled() {
 		Mockito.verifyNoMoreInteractions(validationListener2);
 	}
 
+    @Test
```
nenaraab, line 151: maybe you like to add another test case with ias token and withEnvironmentVariable("IAS_XSUAA_XCHANGE_ENABLED", "false")

nenaraab (commented):
Hi @liga-oz thanks a lot. looks pretty good!
Pls also consider the comments, I've saved them as comments but not in context of the review.

java-security/src/main/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticator.java
liga-oz: @nenaraab , if I do that then it won't do any token verification in case ias to xsuaa token xchange is disabled

java-security/src/test/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticatorTest.java
liga-oz: good point 👍 thanks

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
```diff
@@ -154,8 +154,19 @@ protected TokenAuthenticationResult authenticated(Token token) {
 		return TokenAuthenticatorResult.createAuthenticated(Collections.emptyList(), token);
 	}
 
-	private boolean headerIsAvailable(String authorizationHeader) {
+	boolean headerIsAvailable(String authorizationHeader) {
 		return authorizationHeader != null && !authorizationHeader.isEmpty();
 	}
 
+	TokenAuthenticationResult tokenValidationResult(Token token){
```
nenaraab, line 161: this can be used within the same class
https://github.com/SAP/cloud-security-xsuaa-integration/blob/08abdd4a0b2dfe5bcca8cfab6e76ca5c415525e1/java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java#L43

nenaraab (changes requested):
Hi @liga-oz 
have you applied the formatter?
Thanks,
:-)

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
liga-oz: @nenaraab now I have, thanks :)

nenaraab (approved):

Issue #414, opened 2020-11-10
[spring-xsuaa] Resolve issues with Nimbus-JOSE-JWT version 9.0 com.nimbusds.jose.Header.toJSONObject usage


liga-oz:
With Nimbus-Jose-JWT **version 9.0** method `com.nimbusds.jose.Header.toJSONObject()` signature changes - method returns  Map<String, Object> instead of net.minidev.json.JSONObject as per [changelog](https://bitbucket.org/connect2id/nimbus-jose-jwt/src/24d07948e506fd6e486a488fb0fbd00982cd1cc1/CHANGELOG.txt#lines-1174).

As of Spring Security OAuth2 JOSE version 5.4.0 Nimbus-Jose-JWT 9.0 is used.

Find usages in spring xsuaa modules and resolve potential `java.lang.NoSuchMethodError` errors.

nenaraab:
Related References:
- https://github.com/spring-projects/spring-security/issues/9120
- https://stackoverflow.com/questions/64764311/dependency-conflicts-between-sap-cloud-sdk-and-sap-cloud-security-xsuaa-integrat

Sachpat:
Hi @liga-oz @nenaraab this is one of the questions from our secBOM users @patrickhuy ,  Do you know whether it's possible to use spring boot 2.4.x and downgrade spring security and jose-jwt and then use the xsuaa lib?

liga-oz:
Hi @Sachpat,

if you specify explicitly the jose-jwt library like [here](https://github.com/SAP/cloud-security-xsuaa-integration/issues/413#issuecomment-724712041) and no other methods you use require the newer version of jose-jwt it should be possible.

Best Regards,
Liga

Sachpat:
Thanks @liga-oz . I hope that answers your question @patrickhuy . :)

anshulsaxena-93:
Hi @liga-oz 

Below is my POM still same issue.

Can you please help here?

<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<parent>
		<groupId>org.springframework.boot</groupId>
[...]

PR #432, opened 2020-12-14
fix compatibility with nimbus-jose-jwt version < 9.0 #414

liga-oz:


nenaraab:
Hi @liga-oz

can you please analyze the IT tests:
https://github.com/SAP/cloud-security-xsuaa-integration/pull/432/checks?check_run_id=1552396992

Caused by: java.lang.ClassNotFoundException: com.nimbusds.jose.shaded.json.parser.ParseException
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382) ~[na:1.8.0_275]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:419) ~[na:1.8.0_275]
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352) ~[na:1.8.0_275]
	at java.lang.ClassLoader.loadClass(ClassLoader.java:352) ~[na:1.8.0_275]
	... 73 common frames omitted

Do you see a way to change it without incompatibilities to our consumers? I think we can not force them to upgrade to the latest Spring versions...

nenaraab (changes requested):
as just discussed, we will decide for org.json:json:jar:20200518:compile instead of using nimbus json parser.

spring-xsuaa-test/src/test/java/com/sap/cloud/security/xsuaa/test/JwtGeneratorTest.java
```diff
@@ -77,8 +75,8 @@ public void testTokenWithAttributes() {
 				.addAttribute(ANOTHER_ATTRIBUTE, new String[] { ANOTHER_ATTRIBUTE_VALUE, ANOTHER_ATTRIBUTE_VALUE_2 })
 				.getToken();
 		Map<String, Object> attributes = jwt.getClaimAsMap("xs.user.attributes");
-		assertThat((JSONArray) attributes.get(DUMMY_ATTRIBUTE), contains(DUMMY_ATTRIBUTE));
-		assertThat((JSONArray) attributes.get(ANOTHER_ATTRIBUTE),
+		assertThat(new JSONArray((ArrayList) attributes.get(DUMMY_ATTRIBUTE)), contains(DUMMY_ATTRIBUTE));
```
nenaraab, line 78: In case one would still use minidev lib, then it would still work, right?

nenaraab (approved):
LVGTM!!!

spring-xsuaa-test/src/test/java/com/sap/cloud/security/xsuaa/test/JwtGeneratorTest.java
liga-oz: I would expect it to work, yes.

PR #438, opened 2020-12-23
add explicit java-api dependency in spring xsuaa

liga-oz:


spring-xsuaa/pom.xml
```diff
@@ -48,6 +48,10 @@
 			<groupId>${project.groupId}</groupId>
 			<artifactId>api</artifactId>
 		</dependency>
+		<dependency>
```
nenaraab, line 51: as of now, we can not improve the spring-xsuaa code, as the Token creation logic is not yet available in java-api. That's why I would undo that change.

nenaraab (changes requested):
Hi @liga-oz 
could you please rebase your commit on master... right now it seems that the PR consists of 25 commits ;-)
One little comment and you're done!

Issue #430, opened 2020-12-11
Missing POM dependency for java-api in spring-xsuaa/pom.xml


stevschm:
Hi colleagues,
in spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/XsuaaToken.java you import com.sap.cloud.security.token.InvalidTokenException;

Therefore, there should be a dependency in your spring-xsuaa/pom.xml for the java-api artifactID.

nenaraab:
Hi @stevschm 

thanks for reporting that.

Best regards,
Nena

liga-oz:
Hi @stevschm,
the exception class `com.sap.cloud.security.token.InvalidTokenException` has been replaced by `org.springframework.security.oauth2.server.resource.InvalidBearerTokenException`, if you do not use token exchange functionality, you shouldn't need java-api dependency at all. The changes will be available in the next release of 2.8.1 version.

Best Regards,
Liga

PR #440, opened 2020-12-23
add java-api dependency to spring-xsuaa

liga-oz:
support usage of com.sap.cloud.security.token.InvalidTokenException without token-client module #430

### DIFF
--- a/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
+++ b/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
@@ -16,6 +16,7 @@
 package sample.spring.xsuaa;
 
 import com.sap.cloud.security.xsuaa.XsuaaServiceConfiguration;
+import com.sap.cloud.security.xsuaa.extractor.IasXsuaaExchangeBroker;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.convert.converter.Converter;
@@ -53,6 +54,7 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
 				.anyRequest().denyAll()
 			.and()
 				.oauth2ResourceServer()
+				.bearerTokenResolver(new IasXsuaaExchangeBroker(xsuaaServiceConfiguration))
 				.jwt()
 				.jwtAuthenticationConverter(getJwtAuthenticationConverter());
 		// @formatter:on
--- a/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
+++ b/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
@@ -16,6 +16,7 @@
 package sample.spring.xsuaa;
 
 import com.sap.cloud.security.xsuaa.XsuaaServiceConfiguration;
+import com.sap.cloud.security.xsuaa.client.XsuaaOAuth2TokenService;
 import com.sap.cloud.security.xsuaa.extractor.IasXsuaaExchangeBroker;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Configuration;
@@ -54,7 +55,7 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
 				.anyRequest().denyAll()
 			.and()
 				.oauth2ResourceServer()
-				.bearerTokenResolver(new IasXsuaaExchangeBroker(xsuaaServiceConfiguration))
+				.bearerTokenResolver(new IasXsuaaExchangeBroker(xsuaaServiceConfiguration, new XsuaaOAuth2TokenService()))
 				.jwt()
 				.jwtAuthenticationConverter(getJwtAuthenticationConverter());
 		// @formatter:on
--- a/spring-xsuaa-it/src/test/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBrokerTest.java
+++ b/spring-xsuaa-it/src/test/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBrokerTest.java
@@ -22,20 +22,20 @@ public class IasXsuaaExchangeBrokerTest {
 	private OAuth2TokenService oAuth2TokenService;
 
 	private MockHttpServletRequest request;
-	private static IasXsuaaExchangeBroker tokenXchangeBroker;
+	private static IasXsuaaExchangeBroker cut;
 	private static final String XSUAA_TOKEN = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJleHRfYXR0ciI6eyJlbmhhbmNlciI6IlhTVUFBIn19._cocFCqqATDXx6eBUoF22W9F8VwUVYY59XdLGdEDFso";
 	private static final String IAS_TOKEN = "eyJhbGciOiJIUzI1NiJ9.eyJjbGllbnRfaWQiOiJzYi1qYXZhLWhlbGxvLXdvcmxkIiwiY2lkIjoic2ItamF2YS1oZWxsby13b3JsZCIsImF6cCI6InNiLWphdmEtaGVsbG8td29ybGQiLCJncmFudF90eXBlIjoiY2xpZW50X2NyZWRlbnRpYWxzIiwidXNlcl9pZCI6IjEwMDIxOTEiLCJ1c2VyX25hbWUiOiJXT0xGR0FORyIsImVtYWlsIjoiV09MRkdBTkdAdW5rbm93biIsImlhdCI6MTQ0MjkxMjI0NCwiZXhwIjoxNDQyOTU1MzIyLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvdWFhL29hdXRoL3Rva2VuIiwiZXh0X2F0dHIiOnsic2VydmljZWluc3RhbmNlaWQiOiJhYmNkMTIzNCJ9LCJ6aWQiOiJ6b25lSWQifQ.uzdP2WyfullaZ_5aSjMr2shTfyqqXxbfgShjtfPjigY";
 
 	@Before
 	public void setUp() {
 		request = new MockHttpServletRequest();
-		tokenXchangeBroker = new IasXsuaaExchangeBroker(getXsuaaServiceConfiguration(), oAuth2TokenService);
+		cut = new IasXsuaaExchangeBroker(getXsuaaServiceConfiguration(), oAuth2TokenService);
 	}
 
 	@Test
 	public void xsuaaTokenResolutionTest() {
 		request.addHeader("Authorization", "bearer " + XSUAA_TOKEN);
-		String token = tokenXchangeBroker.resolve(request);
+		String token = cut.resolve(request);
 
 		assertThat(token).isEqualTo(XSUAA_TOKEN);
 	}
@@ -43,7 +43,7 @@ public class IasXsuaaExchangeBrokerTest {
 	@Test
 	public void iasTokenResolutionTest() {
 		request.addHeader("Authorization", "bearer " + IAS_TOKEN);
-		String token = tokenXchangeBroker.resolve(request);
+		String token = cut.resolve(request);
 
 		assertThat(token).isEqualTo(XSUAA_TOKEN);
 	}
@@ -51,13 +51,13 @@ public class IasXsuaaExchangeBrokerTest {
 	@Test(expected = InvalidBearerTokenException.class)
 	public void invalidAuthorizationHeaderTest() {
 		request.addHeader("Authorization", IAS_TOKEN);
-		tokenXchangeBroker.resolve(request);
+		cut.resolve(request);
 	}
 
 	@Test(expected = InvalidBearerTokenException.class)
 	public void invalidAuthorizationHeader2Test() {
 		request.addHeader("Auth", "bearer " + IAS_TOKEN);
-		tokenXchangeBroker.resolve(request);
+		cut.resolve(request);
 	}
 
 	private XsuaaServiceConfiguration getXsuaaServiceConfiguration() {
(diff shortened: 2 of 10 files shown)


nenaraab (commented):
Oh yes, just forgot that... could you please check once more, at which places (beside of XsuaaToken) spring-xsuaa needs the reference to java-api...

because if think in xsuaatoken we should raise a different Spring exception
https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/XsuaaToken.java#L161

For example (didn't checked them):
- org.springframework.security.oauth2.provider.NoSuchClientException
- spring.framework.security...AccessDeniedException
- ???

liga-oz:
Hi @nenaraab,

with the new token xchange implementation it uses many more objects from java-api in `TokenBrokerResolver`, `IasXsuaaExchangeBroker` and `IasToken` classes, for example:

- `com.sap.cloud.security.json.JsonObject`
- `com.sap.cloud.security.config.Service`
- `com.sap.cloud.security.token.Token`
- `com.sap.cloud.security.token.TokenClaims`

Instead of the `InvalidTokenException` could throw 

- `org.springframework.security.oauth2.server.resource.InvalidBearerTokenException` 
- `org.springframework.security.authentication.InsufficientAuthenticationException`

nenaraab:
yes, `InvalidBearerTokenException` would be my favorite!

spring-xsuaa/README.md
```diff
@@ -145,6 +145,38 @@ In detail `com.sap.cloud.security.xsuaa.token.SpringSecurityContext` wraps the S
...
+http.authorizeRequests()
+      .antMatchers("/secured/path/**").hasAuthority("application.Read")
+      .anyRequest().denyAll()
+      .and().oauth2ResourceServer()
+      .bearerTokenResolver(new TokenBrokerResolver(xsuaaServiceConfiguration, cache,
+                                 authenticationMethod));
+```
+where:
+- xsuaaServiceConfiguration: configuration properties from environment
+- cache: application's token-cache
```
nenaraab, line 171: is cache optional? In any case further details might be required: code, link to token-client lib doc?

nenaraab (changes requested):

spring-xsuaa/README.md
liga-oz: doesn't seem to be optional and is used only within the TokenBrokerResolver class directly.

spring-xsuaa/README.md
```diff
@@ -145,6 +145,35 @@ In detail `com.sap.cloud.security.xsuaa.token.SpringSecurityContext` wraps the S
...
+- authenticationMethod: supported authentication method that's being used.
+Currently, the following [authentication methods](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/AuthenticationMethod.java#L3) are supported:
+    - `AuthenticationMethod.BASIC`
+    - `AuthenticationMethod.CLIENT_CREDENTIALS`
+
+Please check out also our [Spring Xsuaa sample](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java) for `TokenBrokerResolver` usage with Basic Authentication method.
+
+To enable token exchange between IAS and XSUAA system environment variable 'IAS_XSUAA_XCHANGE_ENABLED' needs to be provided and enabled as described in [here](https://github.com/SAP/cloud-security-xsuaa-integration/blob/c5b601521c62ad124cd4d5acaf113d93bf265e8a/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java#L73)
```
nenaraab, line 175: I wouldn't link to java-doc (code line). as this is not a stable one...
And i do not know, where to find the official public javadoc :-)

spring-xsuaa/README.md
nenaraab: I've checked the code and the constructors once more. So, the cache is only required in case of authentication method BASIC or CLIENT_CREDENTIALS... but not for pure OAuth... maybe we need a new/separate TokenBrokerResolverDefault implementation, that gets used within the existing one?

nenaraab (changes requested):

spring-xsuaa/README.md
liga-oz: if separate implementation of BearerTokenResolver class, then application developer needs to choose which TokenBrokerResolver to use in securityConfiguration, if that's fine and you think it doesn't complicate things then I can modify the existing IasXsuaaExchangeBroker class to implement the BearerTokenResolver, but that would not be seamless IAS Xsuaa exchange implementation, I guess?

spring-xsuaa/README.md
```diff
@@ -145,6 +145,32 @@ In detail `com.sap.cloud.security.xsuaa.token.SpringSecurityContext` wraps the S
...
+http.authorizeRequests()
+      .antMatchers("/secured/path/**").hasAuthority("application.Read")
+      .anyRequest().denyAll()
+      .and().oauth2ResourceServer()
+      .bearerTokenResolver(new IasXsuaaExchangeBroker(xsuaaServiceConfiguration));
+```
+where:
+- xsuaaServiceConfiguration: configuration properties from environment
+
+To enable token exchange between IAS and XSUAA system environment variable 'IAS_XSUAA_XCHANGE_ENABLED' needs to be provided and enabled. To enable the exchange set the value to any value except 'false' or empty. The exchange between IAS and Xsuaa is disabled by default.
```
nenaraab, line 167: this should come as second step.
1. step required maven dependencies
2. env variable 
3. nothing in case TokenBrokerResolver is already used or  IasXsuaaExchangeBroker incl code snippet
4. sample

spring-xsuaa-it/src/test/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBrokerTest.java
```diff
@@ -0,0 +1,73 @@
...
+	public void xsuaaTokenResolutionTest() {
+		request.addHeader("Authorization", "bearer " + XSUAA_TOKEN);
+		String token = tokenXchangeBroker.resolve(request);
+
+		assertThat(token).isEqualTo(XSUAA_TOKEN);
+	}
+
+	@Test
+	public void iasTokenResolutionTest() {
+
```
nenaraab, line 45: can be removed

spring-xsuaa-it/src/test/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBrokerTest.java
```diff
@@ -0,0 +1,73 @@
...
+	private MockHttpServletRequest request;
+	private static IasXsuaaExchangeBroker tokenXchangeBroker;
+	private static final String XSUAA_TOKEN = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJleHRfYXR0ciI6eyJlbmhhbmNlciI6IlhTVUFBIn19._cocFCqqATDXx6eBUoF22W9F8VwUVYY59XdLGdEDFso";
+	private static final String IAS_TOKEN = "eyJhbGciOiJIUzI1NiJ9.eyJjbGllbnRfaWQiOiJzYi1qYXZhLWhlbGxvLXdvcmxkIiwiY2lkIjoic2ItamF2YS1oZWxsby13b3JsZCIsImF6cCI6InNiLWphdmEtaGVsbG8td29ybGQiLCJncmFudF90eXBlIjoiY2xpZW50X2NyZWRlbnRpYWxzIiwidXNlcl9pZCI6IjEwMDIxOTEiLCJ1c2VyX25hbWUiOiJXT0xGR0FORyIsImVtYWlsIjoiV09MRkdBTkdAdW5rbm93biIsImlhdCI6MTQ0MjkxMjI0NCwiZXhwIjoxNDQyOTU1MzIyLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvdWFhL29hdXRoL3Rva2VuIiwiZXh0X2F0dHIiOnsic2VydmljZWluc3RhbmNlaWQiOiJhYmNkMTIzNCJ9LCJ6aWQiOiJ6b25lSWQifQ.uzdP2WyfullaZ_5aSjMr2shTfyqqXxbfgShjtfPjigY";
+
+	@Before
+	public void setUp() {
+		request = new MockHttpServletRequest();
+		tokenXchangeBroker = new IasXsuaaExchangeBroker(getXsuaaServiceConfiguration(), oAuth2TokenService);
```
nenaraab, line 32: pls rename "tokenXchangeBroker" to "cut" (class under test)

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/TokenBrokerResolver.java
```diff
@@ -186,17 +185,13 @@ private String getBrokerToken(AuthenticationMethod credentialType, String authHe
 			if (oAuth2token == null) {
 				break;
 			}
-			if (iasXsuaaExchangeBroker.isXsuaaToken(oAuth2token)
-					|| !iasXsuaaExchangeBroker.isIasXsuaaXchangeEnabled()) {
+			if (TokenUtil.isXsuaaToken(oAuth2token)
```
nenaraab, line 188: i would just change the order... and perform the check as following:

```java
if (TokenUtil.isIasToXsuaaXchangeEnabled() && !TokenUtil.isXsuaaToken(oAuth2token))  
    // perform token exchange
else //default  
  return oAuth2token;
```

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
```diff
@@ -1,55 +1,73 @@
...
-	 * Verifies if the provided token is Xsuaa token
-	 *
-	 * @param encodedJwtToken
-	 *            Encoded token to be checked
-	 * @return true if provided token is a XSUAA token
-	 */
-	public boolean isXsuaaToken(String encodedJwtToken) {
-		String claims = Base64JwtDecoder.getInstance().decode(encodedJwtToken).getPayload();
+	public IasXsuaaExchangeBroker(XsuaaServiceConfiguration configuration) {
+		this.xsuaaTokenFlows = new XsuaaTokenFlows(
```
nenaraab, line 41: - you can reuse the 3rd constructor by just providing the tokenService
- i would prefer as default tokenService : new DefaultOAuth2TokenService()

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
```diff
@@ -1,55 +1,73 @@
...
+				new XsuaaOAuth2TokenService(new RestTemplate()),
+				new XsuaaDefaultEndpoints(configuration.getUaaUrl()),
+				new ClientCredentials(configuration.getClientId(), configuration.getClientSecret()));
+	}
+
+	public IasXsuaaExchangeBroker(XsuaaServiceConfiguration configuration, OAuth2TokenService tokenService) {
```
nenaraab, line 47: great -> this gives all options to configure for example the rest client or the cache...

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
```diff
@@ -1,55 +1,73 @@
...
-			JSONObject externalAttributeClaim = new JSONObject(claims)
-					.getJSONObject(TokenClaims.XSUAA.EXTERNAL_ATTRIBUTE);
-			String externalAttributeValue = externalAttributeClaim
-					.getString(EXTERNAL_ATTRIBUTE_ENHANCER);
-			return externalAttributeValue.equalsIgnoreCase("xsuaa");
-		} catch (JSONException e) {
-			return false;
+			String oAuth2Token = extractTokenFromRequest(request);
+
+			if (TokenUtil.isXsuaaToken(oAuth2Token)
```
nenaraab, line 60: same like in the other comment, see TokenBrokerResolver

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
```diff
@@ -70,27 +88,28 @@ public String getXsuaaToken(Token iasToken) {
...
+	 * 
+	 * @param oAuth2Token
+	 *            encoded token
+	 * @return IasToken class
+	 * @throws ParseException
+	 *             if provided Jwt token couldn't be parsed
 	 */
-	public boolean isIasXsuaaXchangeEnabled() {
-		return isIasXsuaaXchangeEnabled;
+	Token decodeToken(String oAuth2Token) throws ParseException {
```
nenaraab, line 99: maybe not needed? decodeToken and doIasXsuaaXchange comes always together? Furthermore is decodeToken not the right naming ;-) otherwise i would expected a return type DecodedJwt...

formerly we wanted to separate it to decode the token only once: for the isXsuaaToken check and for the token-exchange itself...

nenaraab (changes requested):
Hi @liga-oz 
looks nice :-) only some minor remarks...

samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
```diff
@@ -53,6 +55,7 @@ protected void configure(HttpSecurity http) throws Exception {
 				.anyRequest().denyAll()
 			.and()
 				.oauth2ResourceServer()
+				.bearerTokenResolver(new IasXsuaaExchangeBroker(xsuaaServiceConfiguration, new XsuaaOAuth2TokenService()))
```
nenaraab, line 58: I would keep the sample with the xsuaaServiceConfiguration parameter (most simple use case)

nenaraab (approved):

samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/SecurityConfiguration.java
liga-oz: with the DefaultOAuth2TokenService it fails to load application context 😕

nenaraab: ahhh... okay because of missing Apache Rest Client... BUT then, we have to adapt our default as well, and then again its not needed...

nenaraab: Sorry, already merged your PR... could you do that in another PR?

PR #448, opened 2021-01-05
add docu describing token exchange usage

liga-oz:


nenaraab (changes requested):
Hi Liga,
reads very good!
I think, step 3 is not required... and then you can (if you like) reduce the Steps to the env variable setting (1) only...
Best regards,
Nena

liga-oz:
> Hi Liga,
> reads very good!
> I think, step 3 is not required... and then you can (if you like) reduce the Steps to the env variable setting (1) only...
> Best regards,
> Nena

@nenaraab I think it is needed, because users might expect that no-arg constructor can also support the token-exchange, which of course is not the case and then they will run into null pointer exception.

nenaraab:
ahh.. i see - you are right (again the doc helped me to see, what I've overseen in the code review)...
- another constructor in XsuaaTokenAuthenticator is hopefully not required, because we have this environment variable
- instead of constructor, i would suggest a private getExchangeBroker() method that initialize the exchangeBroker lazily (when not yet done).... 
- user getter instead of direct access of member exchangeBroker

nenaraab (approved):

PR #450, opened 2021-01-05
[spring-security] Hybrid JWT Decoder

nenaraab:
TODOs:
- [x] depends on https://github.com/SAP/cloud-security-xsuaa-integration/pull/458 and hard code xsuaaAppId to "xsapp"
- [x] depends on https://github.com/SAP/cloud-security-xsuaa-integration/pull/457 => not an issue any longer, as I've decoupled the PRs


spring-security/src/main/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationProperties.java
```diff
@@ -0,0 +1,151 @@
...
+ * </ul>
+ * You can define your own prefix when declaring {@code @ConfigurationProperties("xsuaa")}.
+ * This then needs to be mapped to {@code $vcap.services.<your service instance name>.credentials}
+ * as part of your application properties. <br>
+ *
+ * Alternatively, its possible to directly map the properties to {@code VCAP_SERVICES}
+ * system environment variable: <br>
+ * {@code @ConfigurationProperties("vcap.services.<your service instance name>.credentials")}
+ */
+public class OAuth2ServiceConfigurationProperties implements OAuth2ServiceConfiguration {
```
beckermarc, line 26: Maybe you remember that in CAP we have an abstraction of OSB API Service Bindings. We can meanwhile fill those either from default-env.json, VCAP_SERVICES or from some file-system structure on Kubernetes and with that make the rest of the CAP framework mostly CF/K8s independent.
With the current XSUAA Spring Library we implemented our own XsuaaPropertySourceFactory to connect our ServiceBindings abstractions with your library. I think we should check that this also works with this new configuration approach.

spring-security/src/main/java/com/sap/cloud/security/spring/token/SpringSecurityContext.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * This is an alternative way of accessing jwt tokens of type {@link Token} or
+ * {@link AccessToken} in context of an application using
+ * spring-security-oauth2.
+ *
+ * It uses the {@link SecurityContextHolder} to access Spring's
+ * {@link SecurityContext} and can therefore used also in context of
+ * asynchronous threads.
+ */
+public class SpringSecurityContext {
```
beckermarc, line 28: Will a representation of the token also be available via the spring-independent [SecurityContext](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java) when using this library? This would be great, as it would help libraries to retrieve the token with a single uniform API, independent if the application is using Spring or not.

beckermarc (commented):
Great stuff! This looks promising for providing an even better integration of XSUAA into CAP. I am eager to give it a try if I have some spare time :)

spring-security/src/main/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationProperties.java
nenaraab: Hi @beckermarc thanks for the prompt feedback!
Eager to learn how you did the integration stuff with your own XsuaaPropertySourceFactory implementation. As of now, i have not chosen that in order to support multiple xsuaa bindings, but maybe you have a solution for that too?

spring-security/src/main/java/com/sap/cloud/security/spring/token/SpringSecurityContext.java
nenaraab: not yet but good point - maybe this can be solved with a custom https://docs.spring.io/spring-security/site/docs/4.0.x/apidocs/org/springframework/security/core/context/SecurityContextHolderStrategy.html

nenaraab: see here:
https://github.com/SAP/cloud-security-xsuaa-integration/pull/450/commits/6b601ebbbd7925af4523842574fa2adbff57866e

and here: https://jtrack.wdf.sap.corp/browse/SECAUTH-1270

beckermarc: Looks good! I think auto-configuration for this would be very helpful :)

nenaraab: Done with that: https://github.com/SAP/cloud-security-xsuaa-integration/pull/450/commits/14a22ab7ee880555a252a13d594cd2cafe12be43

spring-security/src/main/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationProperties.java
nenaraab: created a BLI: SECAUTH-1269

nenaraab: Hi @beckermarc 

@PropertySource integration is supported now, as documented here:
https://github.com/SAP/cloud-security-xsuaa-integration/tree/hybrid-jwt-decoder/spring-security#setup-spring-security-oauth-20-resource-server

Issue #452, opened 2021-01-08
Add getSubAccountId() to AccessToken interface


mtolksdorf:
For scenarios requiring the subaccountId to work properly (e.g. OnPremise Connectivty), there is an API getSubAccountId()  in the implementation class XsuaaToken. However, it is not part of the API of the interface AccessToken and if an application likes to stick to the interface so that it does not need to rely on and cast to the implementation, this is currently not possible. 
Proposal add the API and if there is no subaccountId available in the token simply return null or empty String

nenaraab:
Hi @mtolksdorf thanks for the feature reqquest,

we will prove once more and will provide you a fix as soon as possible.

Best regards,
Nena

nenaraab:
available with version 2.8.3

PR #453, opened 2021-01-11
[java-api] AccessToken exposes getSubaccountId() method #452

nenaraab:


java-api/src/main/java/com/sap/cloud/security/token/AccessToken.java
```diff
@@ -50,6 +52,21 @@
 	@Nullable
 	GrantType getGrantType();
 
+	/**
+	 * Returns subaccount identifier. This reflects claim {@code ext_attr.subaccountid} in xsuaa access tokens.
+	 * For example, commercialized multi-tenant applications with a need for metering and billing use
+	 * {@link #getSubaccountId()} method as identifier for the account to be billed.<br>
+	 *
+	 * Multi-tenant applications need to adapt by using the zone ID instead of the subaccount ID as key
```
liga-oz, line 60: I would suggest to remove "by" like this:
Multi-tenant applications need to adapt using the zone ID instead of the subaccount ID as key

liga-oz (changes requested):
minor java doc change, apart from that looks good :)

liga-oz (approved):

PR #458, opened 2021-01-13
Service loader implementation

liga-oz:


java-api/src/main/java/com/sap/cloud/security/token/TokenProvider.java
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	static TokenFactory provider(String providerName) {
+		ServiceLoader<TokenFactory> loader = ServiceLoader.load(TokenFactory.class);
+		for (TokenFactory provider : loader) {
+			if (providerName.equals(provider.getClass().getName())) {
+				return provider;
+			}
+		}
+		throw new ProviderNotFoundException("Token provider " + providerName + " not found");
```
nenaraab, line 23: ... No TokenFactory implementation was found in the class path.

java-security/src/main/java/com/sap/cloud/security/token/HybridTokenFactory.java
```diff
@@ -14,44 +13,33 @@
...
-	 *            the scope converter, e.g. {@link XsuaaScopeConverter}
-	 * @return the new token instance
-	 */
-	public static Token create(String jwtToken, ScopeConverter localScopeConverter) {
+	public Token create(String jwtToken, String xsuaaAppId) {
 		Objects.requireNonNull(jwtToken, "Requires encoded jwtToken to create a Token instance.");
 		DecodedJwt decodedJwt = Base64JwtDecoder.getInstance().decode(removeBearer(jwtToken));
 
 		if (isXsuaaToken(decodedJwt)) {
+			ScopeConverter localScopeConverter = new XsuaaScopeConverter(xsuaaAppId);
```
nenaraab, line 42: Hi @liga-oz 

this change looks now very pretty... beside of this... xsuaa appId is not a runtime information, means it does not change during runtime. That means HybridTokenFactory should create only once the ScopeConverter and for that it need to fetch the appId from the environment (which is cached)...

see also here:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/java-security/src/main/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticator.java#L47

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
```diff
@@ -84,9 +85,14 @@ public AbstractTokenAuthenticator withHttpClient(CloseableHttpClient httpClient)
 	 */
 	public AbstractTokenAuthenticator withServiceConfiguration(OAuth2ServiceConfiguration serviceConfiguration) {
 		this.serviceConfiguration = serviceConfiguration;
+		setupTokenFactory();
 		return this;
 	}
 
+	private void setupTokenFactory() {
+		HybridTokenFactory.withXsuaaAppId(serviceConfiguration.getProperty(CFConstants.XSUAA.APP_ID));
```
nenaraab, line 93: pls check for xsuaa service

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenFactory.java
```diff
@@ -14,49 +19,62 @@
...
 
 		if (isXsuaaToken(decodedJwt)) {
-			return new XsuaaToken(decodedJwt).withScopeConverter(localScopeConverter);
+			return new XsuaaToken(decodedJwt).withScopeConverter(getOrCreateScopeConverter());
 		}
 		return new SapIdToken(decodedJwt);
 	}
 
+	private ScopeConverter getOrCreateScopeConverter() {
+		return new XsuaaScopeConverter(getXsAppId());
```
nenaraab, line 63: pls. keep it as member

nenaraab (changes requested):
very smart solution :-)

nenaraab (approved):

PR #465, opened 2021-01-27
[fix] Change from internal class RSAPublicKeyImpl to RSAPublicKey in JwtGeneratorTest

mwdb:
The test JwtGeneratorTest  was using RSAPublicKeyImpl
As this is an JDK specific internal class, it fails on some JDKs. Fix is to change to the interface.


nenaraab (approved):

nenaraab:
Hi @mwdb thanks a lot for the fix!

PR #467, opened 2021-02-04
performance: make sure ServiceLoader loads services only once

nenaraab:


java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -15,6 +16,12 @@
  * Represents a JSON Web Token (JWT).
  */
 public interface Token extends Serializable {
+	List<TokenFactory> services = new ArrayList() {
+		{
+			ServiceLoader.load(TokenFactory.class).forEach(this::add);
+			LoggerFactory.getLogger(Token.class).info("loaded TokenFactory service providers: {}", this);
+		}
+	};
```
liga-oz, line 24: good idea 💯

liga-oz (approved):
🥇

PR #468, opened 2021-02-05
JavaSecurityContextHolderStrategy and Python Tests

liga-oz:


nenaraab:
hi @liga-oz looks good to me... the only thing, that we have to make sure, that the script can also run (by default) without user-interaction... this needs to be guaranteed.

samples/deploy_and_test.py
```diff
@@ -325,16 +327,18 @@ def test_sayHello_ias(self):
...
+            if(resp.status != 200):
+                logging.warning("In case after adding role collection, user is still not authorized. "
+                                "Check in IAS admin panel that the application's '{}' Subject Name Identifier is set to email. "
+                                "Bug: NGPBUG-139441 "
+                                .format(ias_service.ias_service_name))
+            self.assertEqual(resp.status, 200, 'Expected HTTP status 200')
+            xsappname = self.get_deployed_app().get_credentials_property('xsappname')
+            self.assertRegex(resp.body, xsappname, 'Expected to find xsappname in response')
+        else:
+            logging.warning('test_sayHello_ias was skipped. Enable it with USER_INPUT_ENABLED=true.')
```
liga-oz, line 341: @nenaraab I think it would be also good to mention in warning message to enter y instead of n for the first input. Something like 'test_sayHello_ias was skipped. To run test enable it with USER_INPUT_ENABLED=true and enter 'y' for the user prompt 'Do you want to add role manually?'

nenaraab: yeah.. logging.warning('test_sayHello_ias was skipped. Enable it with USER_INPUT_ENABLED=true.')

nenaraab (approved):

PR #473, opened 2021-02-16
CI and python script should include java-security-sample-ias

nenaraab:
- the service and service key was not successfully deleted

### DIFF
--- a/samples/deploy_and_test.py
+++ b/samples/deploy_and_test.py
@@ -83,11 +83,12 @@ class SampleTest(abc.ABC, unittest.TestCase):
         self.__api_access = None
         self.__ias_access = None
         self.credentials = credentials
-        self.get_app().deploy()
+        self.cf_app = self.get_app()
+        self.cf_app.deploy()
         time.sleep(2)  # waiting for deployed apps to be available
 
     def tearDown(self):
-        self.get_app().delete()
+        self.cf_app.delete()
         if self.__api_access is not None:
             self.__api_access.delete()
         if self.__ias_access is not None:
@@ -124,9 +125,9 @@ class SampleTest(abc.ABC, unittest.TestCase):
 
     def get_deployed_app(self):
         if self.__deployed_app is None:
-            deployed_app = self.cf_apps.app_by_name(self.get_app().name)
+            deployed_app = self.cf_apps.app_by_name(self.cf_app.name)
             if deployed_app is None:
-                logging.error('Could not find app: ' + self.get_app().name)
+                logging.error('Could not find app: ' + self.cf_app.name)
                 exit()
             self.__deployed_app = deployed_app
         return self.__deployed_app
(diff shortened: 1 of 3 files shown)


samples/deploy_and_test.py
```diff
@@ -789,27 +805,27 @@ def __str__(self):
...
-        subprocess.run(
-            ['cf', 'create-service', 'xsuaa', 'application', self.xsuaa_service_name, '-c', 'xs-security.json'],
-            cwd=self.working_dir, stdout=cf_logs, check=True)
+        if self.xsuaa_service_name is not None:
+            logging.info("Creating Xsuaa service '{}'".format(self.xsuaa_service_name))
+            subprocess.run(
+                ['cf', 'create-service', 'xsuaa', 'application', self.xsuaa_service_name, '-c', 'xs-security.json'],
+                cwd=self.working_dir, stdout=cf_logs, check=True)
         if self.identity_service_name is not None:
-            logging.info("Creating IAS service '{}'".format(self.identity_service_name))
```
liga-oz, line 810: this logging output should be added in `__create_ias_service()` method

samples/deploy_and_test.py
```diff
@@ -819,14 +835,16 @@ def deploy(self):
     def delete(self):
         logging.info("Deleting '{}'".format(self.name))
         subprocess.run(['cf', 'delete', '-f', '-r', self.name], stdout=cf_logs)
+        subprocess.run(['cf', 'delete-orphaned-routes', '-f'], stdout=cf_logs)
         if self.app_router_name is not None:
             logging.info("Deleting '{}' app router".format(self.app_router_name))
             subprocess.run(['cf', 'delete', '-f', '-r', self.app_router_name], stdout=cf_logs)
         logging.info("Deleting '{}' Xsuaa service".format(self.xsuaa_service_name))
```
liga-oz, line 842: this logging info should then go also in the if statement L844

samples/deploy_and_test.py
```diff
@@ -498,7 +512,7 @@ def __execute(self, req):
 
 class IasAccess:
     def __init__(self, ias_name):
-        self.ias_service_key_name = "ias-access-key"
+        self.ias_service_key_name = ias_name + '-key'
```
liga-oz, line 515: really good one 🥇

liga-oz (changes requested):
just few minor changes for logging output, other then that very good :)

samples/deploy_and_test.py
nenaraab: not sure, delete-orphaned-routes deletes routes from the application itself...

samples/deploy_and_test.py
nenaraab: yep, done :-)

samples/deploy_and_test.py
```diff
@@ -584,17 +586,12 @@ def fetch_ias_token(self, user):
...
 
     def delete(self):
-        IasAccess.delete(self.ias_service_name)
-
-    @staticmethod
-    def delete(ias_service_name):
-        ias_service_key_name = ias_service_name + '-key'
         logging.info("Deleting service key '{}' for '{}' IAS service"
-                     .format(ias_service_key_name, ias_service_name))
+                     .format(self.ias_service_key_name, self.ias_service_name))
```
nenaraab: Hi @liga-oz is it a good practice to use self in a static method? instead we want to use parameter and local variable...

nenaraab (commented):
hi @liga-oz thanks for the nice fix!

PR #475, opened 2021-02-19
Zone id support in clientCredentialsFlow

liga-oz:


token-client/src/test/java/com/sap/cloud/security/xsuaa/client/AbstractOAuth2TokenServiceTest.java
```diff
@@ -21,6 +21,7 @@
 
 	public static final URI TOKEN_ENDPOINT_URI = URI.create("http://test.token.endpoint/oauth/token");
 	public static final String SUBDOMAIN = "subdomain";
+	public static final String ZONE_ID = "zone";
```
nenaraab, line 24: for compatibility i would preferred ZONE_ID = null.

nenaraab (approved):

PR #479, opened 2021-02-24
[spring-xsuaa] quickfix for supporting junit4 and junit5 with mvn surefire plugin

liga-oz:


nenaraab:
Hi @liga-oz 
wow - that was fast - thanks a lot!!!

nenaraab (approved):
+1

PR #480, opened 2021-03-02
Return the list of audiences that was used for audience validation in JwtAudienceValidator

edmundham:


nenaraab:
Hi @edmundham 
thanks a lot for the PR! I understand that the way you have suggested the logged error message is much more understandable for an application developer :-)
But in case, we get an issue in regard to failed token validation its for us developers much better to have the "audiences" from the token in order to provide further consulting... or to double check our implementation. As its not recommended to log the whole token, this is a good compromise. 

Consequently, I would like to keep it as it is right now.

Thanks and best regards,
Nena

PR #481, opened 2021-03-03
Ignore sb- prefix when checking audience

edmundham:
Currently `sb-` isn't ignored when validating audiences from the token. By looking at the existing unit tests, additional CIDs that doesn't have `sb-` are added to the trusted client ID list, but this behaviour is more intuitive if it's done in `JwtAudienceValidator`.


nenaraab:
Hi @edmundham 
thanks for your pull request. But the decision was to configure the JwtAudienceValidator in case of XSUAA 
- with the client id (which starts with 'sb-' prefix) and
- with the Xsuaa application identifier, which is required in case of clone tokens.

PR #485, opened 2021-03-09
SpringSecurityContext should also support Spring OIDC id tokens

nenaraab:


java-security/src/test/java/com/sap/cloud/security/adapter/spring/SpringOAuth2ClientSecurityContextTest.java
```diff
@@ -0,0 +1,62 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.singleton;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SpringOAuth2ClientSecurityContextTest {
+	SapIdToken sapIdToken;
+
+	@Before
```
liga-oz, line 28: I would suggest for all new test classes that we create to use only junit5 going forward :interrobang:

liga-oz (approved):
Just a suggestion to switch to junit5 tests going forward

PR #489, opened 2021-03-24
Update oidcConfigurationService assertion in JwtSignatureValidator

davidnazw:
should be checking oidcConfigurationService instead of tokenKeyService


liga-oz (approved):
Good finding, thanks a lot @davidnazw 💯

PR #490, opened 2021-03-25
ProofOfPossession X5t validator

liga-oz:


java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -215,6 +215,28 @@ public JwtGenerator withClaimValue(String claimName, JsonObject object) {
...
+	 *
+	 * @param claimName
+	 *            the name of the claim to be set.
+	 * @param object
+	 *            the JSONObject representation of claim value to be set.
+	 * @return the builder object.
+	 * @throws JsonParsingException
+	 *             if the given object does not contain valid json.
+	 */
+	public JwtGenerator withClaimValue(String claimName, JSONObject object) {
```
nenaraab, line 231: why needed, there is already: public JwtGenerator withClaimValue(String claimName, JsonObject object)

java-security/README.md
```diff
@@ -184,6 +184,11 @@ Steps to enable token exchange:
 
 The authenticator is used in the following [sample](/samples/java-security-usage).
 
+### X509 certificate thumbprint validation
+[IasTokenAuthenticator](src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java) supports JWT Certificate Thumbprint Confirmation Method. See specification [here](https://tools.ietf.org/html/rfc8705#section-3.1). 
+This feature is enabled by default. It can be disabled by setting environment variable `X509_THUMBPRINT_CONFIRMATION_ACTIVE` to false.
```
nenaraab, line 189: pls provide also a troubleshoot guide which explains validator and error in more detail....

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -3,17 +3,38 @@
...
 	public Token extractFromHeader(String authorizationHeader) {
 		return new SapIdToken(authorizationHeader);
 	}
 
+	@Override
+	public TokenAuthenticationResult validateRequest(ServletRequest request, ServletResponse response) {
+		HttpServletRequest httpRequest = (HttpServletRequest) request;
+		SecurityContext.setCertificate(x509Selector.getCertificate(httpRequest));
+		try {
+			super.createIasJwtValidators();
```
nenaraab, line 30: see comment above...

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
nenaraab: idea: change parameter to map.

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtCnfValidator.java
```diff
@@ -0,0 +1,83 @@
...
+	 * @param token
+	 *            token to be validated
+	 * @return validation result. Result is valid when both thumbprints match.
+	 */
+	@Override
+	public ValidationResult validate(Token token) {
+
+		String cnf = extractCnfThumbprintFromToken(token);
+		LOGGER.info("Cnf thumbprint: {}", cnf);
+		if (cnf == null && token.getAudiences().size() == 1 && token.getAudiences().contains(trustedClientId)) {
```
nenaraab, line 48: for this check pls move cnf check into the else block... and pls check whether the 3rd check aud == trustedClientId is required (will this validator be executed in case the audience validation (which comes first) failed?

java-security/src/main/java/com/sap/cloud/security/token/x509/CFx509CertSelector.java
```diff
@@ -0,0 +1,24 @@
+package com.sap.cloud.security.token.x509;
```
nenaraab, line 1: not sure about the package

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -62,4 +97,12 @@ public static void clearToken() {
 		}
 	}
 
+	/**
+	 * Clears the current token and certificate from thread wide storage.
+	 */
+	public static void clear() {
```
nenaraab, line 103: pls add @Deprecated annotation to clearToken

java-security/src/test/resources/x509Base64.txt
```diff
@@ -0,0 +1,35 @@
+-----BEGIN CERTIFICATE-----
```
nenaraab, line 1: is it the same formatting like you got it from vcap_services?

token-client/src/main/java/com/sap/cloud/security/xsuaa/mtls/SSLContextFactory.java
```diff
@@ -29,9 +29,8 @@
 /**
  * Creates a SSLContext (without Bouncy Castle crypto lib).
  *
- * @deprecated This class might be removed again in the future
  */
-@Deprecated
+// TODO move to somewhere else?
```
nenaraab, line 33: open TODO

token-client/src/test/java/com/sap/cloud/security/xsuaa/mtls/X509ParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class X509ParserTest {
+
+	private static String x509;
+	private static final String cnf = "fU-XoQlhMTpQsz9ArXl6zHIpMGuRO4ExLKdLRTc5VjM";
+
+	@BeforeAll
+	static void beforeAll() throws IOException {
+		x509 = IOUtils.resourceToString("/x509Base64.txt", StandardCharsets.US_ASCII);
```
nenaraab, line 23: why not UTF_8?

token-client/src/test/resources/x509Base64.txt
```diff
@@ -0,0 +1,35 @@
+-----BEGIN CERTIFICATE-----
```
nenaraab, line 1: is it like the one you get via vcap services? I'm missing \n linebreaks

token-client/src/test/java/com/sap/cloud/security/xsuaa/mtls/X509ParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+	@Test
+	void normalizeThumbprintTest() throws NoSuchAlgorithmException, CertificateException {
+		String normalizedX509 = normalizeThumbprint(getX509Thumbprint(x509));
+		assertFalse(normalizedX509.endsWith("="));
+		assertFalse(normalizedX509.contains("\n"));
+	}
+
+	@Test
+	void getX509ThumbprintTest() throws NoSuchAlgorithmException, CertificateException {
+		assertEquals(cnf, (getX509Thumbprint(x509)));
```
nenaraab, line 35: outer "()" are not required

token-client/src/main/java/com/sap/cloud/security/xsuaa/mtls/X509Parser.java
```diff
@@ -0,0 +1,100 @@
...
+		x509Certificate = decodeX509(certificate);
+		MessageDigest digest = MessageDigest.getInstance("SHA-256");
+		byte[] hashedX509 = digest.digest(x509Certificate.getEncoded());
+		String base64UrlEncoding = normalizeThumbprint(Base64.getUrlEncoder().encodeToString(hashedX509));
+		LOGGER.debug("Base64-url encoded thumbprint: {}", base64UrlEncoding);
+		return base64UrlEncoding;
+	}
+
+	public static String getJwtThumbprint(String base64cert) throws NoSuchAlgorithmException {
+		base64cert = base64cert
```
nenaraab, line 47: still required? Wasn't normalization done by X509Certificate?

java-security/src/main/java/com/sap/cloud/security/token/x509/X509CertSelector.java
```diff
@@ -0,0 +1,11 @@
+package com.sap.cloud.security.token.x509;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Interface for accessing X509 certificates.
+ */
+public interface X509CertSelector {
+
+	String getCertificate(HttpServletRequest request);
```
nenaraab, line 10: maybe we should offer a default create method, that checks whether its in CF environement and return CFx509CertSelector

java-security/src/main/java/com/sap/cloud/security/token/x509/X509CertSelector.java
```diff
@@ -0,0 +1,11 @@
+package com.sap.cloud.security.token.x509;
```
nenaraab, line 1: this schould be moved to java-api.

nenaraab (changes requested):
Hi @liga-oz 
looks good!

- not sure whether sample is performing checks on CF as I haven't seen any usage of X509CertSelector.java to fill the SecurityContext (maybe I've overseen it)?
- before releasing it it might be good i've we check again the package structure and which class should be provided with which component.

Kind regards,
Nena

token-client/src/test/resources/x509Base64.txt
liga-oz: nope, needs to be changed, remove the header and footer as that's how it looks in the x-forwarded-client-cert header

token-client/src/main/java/com/sap/cloud/security/xsuaa/mtls/X509Parser.java
liga-oz: this is the thumbprint generator that doesn't use X509Certificate class, possibly need to remove, or test the performance for both options and then decide

nenaraab: I would keep only one: standard javax.security functionality would be my favorite

token-client/src/test/resources/x509Base64.txt
nenaraab: ah, okay, then maybe it should be renamed to cf-forwarded-client-cert.txt or sth like that

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtCnfValidator.java
liga-oz: if aud is empty check azp with getAudiences method

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
```diff
@@ -137,17 +138,34 @@ public AbstractTokenAuthenticator withValidationListener(ValidationListener vali
...
+			if (isSysEnvPropertyEnabled("X509_THUMBPRINT_CONFIRMATION_ACTIVE", true)) {
+				this.tokenValidator = jwtValidatorBuilder.with(new JwtCnfValidator(serviceConfiguration.getClientId()))
+						.build();
+			} else {
+				this.tokenValidator = jwtValidatorBuilder.build();
+			}
+		}
+	}
+
+	private JwtValidatorBuilder getJwtValidatorBuilder() {
```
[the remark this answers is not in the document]
liga-oz: @nenaraab ja, but I can't access tokenValidator instance variable, then I need to create setter for it.

java-security/src/test/resources/x509Base64.txt
liga-oz: file name and formatting changed according to received cert from x-forwarded-client-cert header

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtCnfValidator.java
liga-oz: @nenaraab  I've checked that and only in case of Xsuaa azp is added to audience list and then validated -> [JwtAudienceValidator](https://github.com/SAP/cloud-security-xsuaa-integration/blob/d1fe8fde82997f505ae6aea4252333c26c82591b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtAudienceValidator.java#L100)

 So are you sure that we can accept azp also in case of IAS? in that case I would adapt the audience validator.

java-security/src/main/java/com/sap/cloud/security/token/x509/CFx509CertSelector.java
liga-oz: agreed to leave it where it is for now

java-security/src/main/java/com/sap/cloud/security/token/x509/X509CertSelector.java
liga-oz: as discussed not moving due to create() method

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -67,4 +105,12 @@ public static void clearToken() {
...
+		clearCertificate();
+		clearToken();
```
nenaraab, line 113: now you are calling a deprecated method...

java-api/src/main/java/com/sap/cloud/security/x509/X509Constants.java
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.sap.cloud.security.x509;
+
+public final class X509Constants {
+
+	private X509Constants() {
+	}
+
+	public static final String X5T_VALIDATOR_ENABLED = "X5T_VALIDATOR_ENABLED";
```
nenaraab, line 13: should not be required in that PR, same is true for ACCEPT_CALLERS_BOUND_TO_SAME_ID_SERVICE

java-api/src/main/java/com/sap/cloud/security/x509/X509Parser.java
```diff
@@ -0,0 +1,132 @@
...
+	/**
+	 * Decodes X509 base64 encoded certificate into X509Certificate class
+	 * 
+	 * @param encodedX509
+	 *            base64 encoded X509 certificate
+	 * @return Decoded X509Certificate
+	 * @throws CertificateException
+	 *             if String value of certificate cannot be parsed
+	 */
+	private static X509Certificate decodeX509(@Nonnull String encodedX509) throws CertificateException {
```
nenaraab, line 94: this method is called parseCertString in the go client lib

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -20,6 +20,41 @@ private SecurityContext() {
 	}
 
 	private static final ThreadLocal<Token> tokenStorage = new ThreadLocal<>();
+	private static final ThreadLocal<String> certificateStorage = new ThreadLocal<>();
```
nenaraab, line 23: in the go client lib I've decided to store not the string but the x509 certificate object. Otherwise the different X509 validators needs to the same multiple ways as all of them require the object and not the string. is that possible in your implementation?

java-api/src/main/java/com/sap/cloud/security/x509/X509Parser.java
```diff
@@ -0,0 +1,132 @@
...
+	 * @see <a href="https://tools.ietf.org/html/rfc8705#section-3.1"></a>
+	 * @param base64certificate
+	 *            certificate encoded in Base64, supports also PEM encoded
+	 *            certificates
+	 * @return Thumbprint of X509 certificate
+	 * @throws NoSuchAlgorithmException
+	 *             when a particular cryptographic algorithm is requested but is not
+	 *             available in the environment
+	 */
+	public static String getCertificateThumbprint(String base64certificate) throws NoSuchAlgorithmException {
```
nenaraab, line 62: for which purpose this method is there? I think this one isn't used and needed?

java-api/src/main/java/com/sap/cloud/security/x509/X509Parser.java
```diff
@@ -0,0 +1,132 @@
...
+	 * @param certificate
+	 *            Base64 encoded X509 certificate
+	 * @return Thumbprint of X509 certificate
+	 * @throws NoSuchAlgorithmException
+	 *             when a particular cryptographic algorithm is requested but is not
+	 *             available in the environment
+	 * @throws CertificateException
+	 *             is thrown if errors occur while decoding X509 certificate
+	 */
+	public static String getX509Thumbprint(String certificate) throws NoSuchAlgorithmException, CertificateException {
```
nenaraab, line 39: here i would change the parameter type to X509Certificate

java-security/README.md
```diff
@@ -179,6 +179,26 @@ For the integration of different Identity Services the [`TokenAuthenticator`](/j
...
 
+
+### ProofOfPossession validation 
+#### X509 certificate thumbprint `X5t` validation
+[JwtX5tValidator](src/main/java/com/sap/cloud/security/token/validation/validators/JwtX5tValidator.java) offers JWT Certificate Thumbprint `X5t` confirmation method's validation. See specification [here](https://tools.ietf.org/html/rfc8705#section-3.1). 
+This validator is not part of the default `CombiningValidator`, it needs to be added manually to `JwtValidatorBuilder` to use it. 
+It can be done in the following manner:
+```java
+JwtValidatorBuilder.getInstance(OAuth2ServiceConfiguration)
+    .with(new JwtX5tValidator())
```
nenaraab, line 190: 💯

java-api/src/main/java/com/sap/cloud/security/token/TokenClaims.java
```diff
@@ -43,8 +43,8 @@ private TokenClaims() {
 	public static final String SAP_GLOBAL_ZONE_ID = "zone_uuid"; // tenant GUID
 	public static final String GROUPS = "groups"; // scim groups
 	public static final String AUTHORIZATION_PARTY = "azp"; // Authorization party contains OAuth client identifier
-	static final String CNF = "cnf"; // X509 certificate ("cnf" (confirmation)) claim
-	static final String CNF_X509_THUMBPRINT = "x5t#S256"; // X509 certificate thumbprint confirmation method
+	public static final String CNF = "cnf"; // X509 certificate ("cnf" (confirmation)) claim
+	public static final String CNF_X509_THUMBPRINT = "x5t#S256"; // X509 certificate thumbprint confirmation method
```
nenaraab, line 47: ->rename to CNF_MEMBER_X5T ?

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
```diff
@@ -143,15 +143,19 @@ public AbstractTokenAuthenticator withValidationListener(ValidationListener vali
...
-			validationListeners.forEach(jwtValidatorBuilder::withValidatorListener);
-			tokenValidator = jwtValidatorBuilder.build();
+		if (this.tokenValidator == null) {
+			this.tokenValidator = getJwtValidatorBuilder().build();
 		}
-		return tokenValidator;
+		return this.tokenValidator;
+	}
+
+	JwtValidatorBuilder getJwtValidatorBuilder() {
```
nenaraab, line 152: any change neside of method extraction?

java-security/src/main/java/com/sap/cloud/security/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,34 @@
...
+	 * false. Any other value is interpreted as true.
+	 *
+	 * @param propertyName
+	 *            name of System Environment variable
+	 * @param defaultEnabled
+	 *            is the default value of this property enabled? true if enabled :
+	 *            false if disabled
+	 * @return boolean value
+	 */
+	public static boolean isSysEnvPropertyEnabled(String propertyName, boolean defaultEnabled) {
```
nenaraab, line 25: ConfigurationUtil is not required/used in this PR

java-security/src/main/java/com/sap/cloud/security/x509/CfX509CertExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+
+import javax.annotation.Nullable;
+import javax.servlet.http.HttpServletRequest;
+
+import static com.sap.cloud.security.x509.X509Constants.FWD_CLIENT_CERT_HEADER;
+
+/**
+ * X509 certificate accessor implementation for Cloud Foundry environment.
+ */
+public class CfX509CertExtractor implements X509CertExtractor {
```
nenaraab, line 14: It should be the same for K8s

java-security/src/main/java/com/sap/cloud/security/x509/K8sX509CertExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+public class K8sX509CertExtractor implements X509CertExtractor {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(K8sX509CertExtractor.class);
+
+	@Override
+	@Nullable
+	public String getClientCertificate(HttpServletRequest request) {
+		String clientCert = request.getHeader(FWD_CLIENT_CERT_HEADER);
+		LOGGER.debug("{} = {}", FWD_CLIENT_CERT_HEADER, clientCert);
+		return X509Parser.formatBase64Cert(clientCert);
```
nenaraab, line 23: maybe this can be (is already) handled in the X509Parser.decodeX509?
I guess then the both classes and the interface is not required any longer.

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
```diff
@@ -220,6 +220,28 @@ public JwtGenerator withClaimValue(String claimName, JsonObject object) {
...
+	 *
+	 * @param claimName
+	 *            the name of the claim to be set.
+	 * @param map
+	 *            map of key value pairs of claims to be set.
+	 * @return the builder object.
+	 * @throws JsonParsingException
+	 *             if the given object does not contain valid json.
+	 */
+	public JwtGenerator withClaimValue(String claimName, Map<String, String> map) {
```
nenaraab, line 236: Thats correct and good to have, but is this used somewhere, e.g. in IasTokenAuthenticatorX509Test.java

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -34,4 +50,13 @@ protected OAuth2ServiceConfiguration getServiceConfiguration() {
 	protected OAuth2ServiceConfiguration getOtherServiceConfiguration() {
 		return null;
 	}
+
+	@Override
+	final Validator<Token> getOrCreateTokenValidator() {
```
nenaraab, line 55: I don't get the gist of this implementation - can you please shortly explain?

java-security/pom.xml
```diff
@@ -93,6 +93,12 @@
 			<artifactId>junit-jupiter-params</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
nenaraab, line 96: I think its not required here...

nenaraab (changes requested):

java-api/src/main/java/com/sap/cloud/security/x509/X509Parser.java
liga-oz: I can rename it to parseCert() but I don't think String is required in the method name

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
liga-oz: If for client certificate validation in certificate comparison only DN needs to match, it makes sense to switch, but if the whole certificate needs to be matched then it's better to store base64 encoded cert.

java-api/src/main/java/com/sap/cloud/security/x509/X509Parser.java
liga-oz: that's described in javadoc, one of the 2 methods can be removed based on the strategy what we store in the  securitycontext -> cert encoded in base64 or X509 cert object

java-security-test/src/main/java/com/sap/cloud/security/test/JwtGenerator.java
liga-oz: there is no other easy way provided in the JwtGenerator to create nested object. I've used it to generate the token with x5t claim for testing resources.

java-security/src/main/java/com/sap/cloud/security/x509/CfX509CertExtractor.java
liga-oz: but there's a difference how x-forwarded-cert can be delivered in k8s env, that's why 2 implementations

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
liga-oz: is there a need to clear the securitycontext i.e. separately for token and for certificate? If not then with the next major release this method needs to become private, if yes then i can de-depracete it.

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
liga-oz: it seems to be leftover for the default validator integration with x509 specific validators -> not for this PR

java-security/src/main/java/com/sap/cloud/security/servlet/AbstractTokenAuthenticator.java
liga-oz: only the return type

java-api/src/main/java/com/sap/cloud/security/x509/X509Constants.java
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.sap.cloud.security.x509;
+
+public final class X509Constants {
+
+	private X509Constants() {
+	}
+
+	public static final String FWD_CLIENT_CERT_HEADER = "x-forwarded-client-cert";
```
nenaraab, line 13: as it is only used in X509CertificateExtractor.java and in tests I would move the constant to that class.

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -8,17 +8,31 @@
...
+import com.sap.cloud.security.x509.X509CertificateExtractor;
 
 import javax.annotation.Nullable;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
 
 public class IasTokenAuthenticator extends AbstractTokenAuthenticator {
 
+	private final X509CertificateExtractor x509CertificateExtractor = X509CertificateExtractor.getInstance();
```
nenaraab, line 22: you can inline that

java-security/src/main/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticator.java
```diff
@@ -32,6 +33,7 @@
 	private static final Logger LOGGER = LoggerFactory.getLogger(XsuaaTokenAuthenticator.class);
 	private static final String IAS_XSUAA_ENABLED = "IAS_XSUAA_XCHANGE_ENABLED";
 	private IasXsuaaExchangeBroker exchangeBroker;
+	private final X509CertificateExtractor x509CertificateExtractor = X509CertificateExtractor.getInstance();
```
nenaraab, line 36: you can inline that one

java-security/src/main/java/com/sap/cloud/security/x509/X509CertificateExtractor.java
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	@Nullable
+	public X509Certificate getClientCertificate(HttpServletRequest request) {
+		String clientCert = request.getHeader(FWD_CLIENT_CERT_HEADER);
+		LOGGER.debug("{} = {}", FWD_CLIENT_CERT_HEADER, clientCert);
+		if (clientCert != null && !clientCert.isEmpty()) {
+			try {
+				return X509Parser.parseCertificate(clientCert);
+			} catch (CertificateException e) {
+				LOGGER.debug("Could not parse the certificate", e);
```
nenaraab, line 49: i would log a warning here

nenaraab (changes requested):
Hi @liga-oz looks very clean and ready. I have only few minor remarks.

java-api/src/main/java/com/sap/cloud/security/x509/X509Constants.java
liga-oz: yes for now, but with the next validators there will be sys env variables that will control the validator usage, I just removed it to make this PR slim

java-security/src/main/java/com/sap/cloud/security/x509/X509CertificateExtractor.java
liga-oz: i didnt want to do that because in case client credentials are used for incoming request the certificate will be null and it would log the warning, even thou the request was successful

nenaraab: I've understood the code in that way, that in case there is no header set, the code block is skipped.

java-api/src/main/java/com/sap/cloud/security/x509/X509Constants.java
nenaraab: ok

java-security/src/main/java/com/sap/cloud/security/x509/X509CertificateExtractor.java
liga-oz: oh yes you're right, forgot that I changed it. Thanks!

nenaraab (approved):
@liga-oz very nice 🥇

PR #492, opened 2021-03-30
enhance getAudience method to use also azp

liga-oz:
enhance getAudience method to use azp in case audience claim is empty


nenaraab:
hi @strehle  what's your opinion - is it okay when Token.getAudiences() returns "azp" in case "aud" is empty? This is also used in context of audience validation.

PR #498, opened 2021-04-07
Fix iasTokenBroker open endpoint bug

liga-oz:


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
```diff
@@ -46,20 +48,22 @@ public IasXsuaaExchangeBroker(XsuaaServiceConfiguration configuration) {
 	@Override
 	@Nullable
 	public String resolve(HttpServletRequest request) {
+		String oAuth2Token = extractTokenFromRequest(request);
+		if (oAuth2Token == null) {
+			logger.error("Request did not have Authorization header containing bearer token.");
```
nenaraab, line 53: Please do not log as error, otherwise we will get a lot of "false-positive" tickets from apps that use /health endpoint for e.g. availability monitoring...

In this case I would just log an info like, "skip token exchange as no token was provided via authorization header"

nenaraab (changes requested):
Only little request in regard to logging - the rest is perfect 💯

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/extractor/IasXsuaaExchangeBroker.java
liga-oz: done

nenaraab (approved):

PR #502, opened 2021-04-13
upgrade spring-security version to 5.4.6

liga-oz:
spring-security 5.4.5 has a vulnerability https://github.com/spring-projects/spring-security/issues/4001 that is fixed in the 5.4.6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 All notable changes to this project will be documented in this file.
-
+## 2.8.11 and 0.1.5 [BETA]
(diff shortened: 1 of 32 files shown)


artem-smotrakov:
This vulnerability doesn't seem to be fixed in `5.4.6`. This pull requests address it but it is not merged yet

https://github.com/spring-projects/spring-security/pull/8082

nenaraab:
Hi @artem-smotrakov thanks for the link! 
My favorite would be to wait for a patched spring-boot version to solve the transient dependencies of our spring starter projects.
@liga-oz would you mind to open an issue to track that?

PR #505, opened 2021-04-21
OAuth2ServiceException provides getStatusCode()

nenaraab:


token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2ServiceException.java
```diff
@@ -12,15 +12,31 @@
 public class OAuth2ServiceException extends IOException {
 
 	private static final long serialVersionUID = 1L;
+	private Integer statusCode = 0;
```
liga-oz, line 15: I would suggest to rename statusCode to httpStatusCode so that it is 100% clear what kinf status code it is

liga-oz (approved):
Just one suggestion to rename the statusCode to httpStatusCode, but I leave it up to you to decide 😉

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2ServiceException.java
nenaraab: 👍

liga-oz (approved):
Great 💯

PR #511, opened 2021-05-07
Enhance x509 constants

liga-oz:


java-api/src/main/java/com/sap/cloud/security/token/TokenClaims.java
```diff
@@ -28,6 +28,8 @@ private TokenClaims() {
 	public static final String SAP_GLOBAL_ZONE_ID = "zone_uuid"; // tenant GUID
 	public static final String GROUPS = "groups"; // scim groups
 	public static final String AUTHORIZATION_PARTY = "azp"; // Authorization party contains OAuth client identifier
+	public static final String CNF = "cnf"; //X509 certificate ("cnf" (confirmation)) claim
+	public static final String X509_THUMBPRINT = "x5t#S256"; //X509 certificate thumbprint confirmation method
```
nenaraab, line 32: Hi @liga-oz i guess its a claim value? In that case I would't add it here...

nenaraab (approved):

PR #536, opened 2021-06-15
XsuaaJwtDecoder raises BadJwtException instead of JwtException

nenaraab:
As of org.springframework.security version 5.5.0 AuthenticationServiceException gets no longer handled and does not result into 401 status code. 

But BadJwtException results into InvalidBearerTokenException which then gets mapped to 401 status code as expected.

Further References
- https://github.com/spring-projects/spring-security/issues/9395
- https://github.com/spring-projects/spring-security/commit/0c3754c81133afd33bfdac28569473d8ac2cdaff#diff-eff4fee1a4b87fa8efaa934f4d2383df1c50192c36870d6b51db3bd7ff37d35e


nenaraab:
stakeholder tested successfully

PR #539, opened 2021-06-18
Move to SAP-internal KnowledgeBase

nenaraab:
Hi @liga-oz
as soon as you've moved it to KnowledgeBase -> 03_ApplicationSecurity we can remove it from here.


liga-oz:
@nenaraab, done 👍🏻

liga-oz (approved):

PR #540, opened 2021-06-18
[java-security] Support ias custom domain using `ias_iss` claim

nenaraab:
- removes usage of url in ValidatorBuilder which was required until `domains` was provided with identity service binding.
- in `ias_iss` is given and not empty, it checks whether its a valid url and checks whether this matches one of the valid domains of the identity service. The check whether `ias` matches to any given domains is skipped in that case.


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtIssuerValidator.java
```diff
@@ -77,39 +61,69 @@
...
 	 */
-
 	@Override
 	public ValidationResult validate(Token token) {
-		String issuer = token.getClaimAsString(TokenClaims.ISSUER);
-		if (issuer == null || issuer.trim().isEmpty()) {
-			return createInvalid(
-					"Issuer validation can not be performed because Jwt token does not contain 'iss' claim.");
+		ValidationResult validationResult;
+		String ias_issuer_url = token.getClaimAsString(IAS_ISSUER);
```
liga-oz, line 67: Is there a reason why you don't use camel case for a local variable?

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtIssuerValidator.java
```diff
@@ -77,39 +61,69 @@
 	 * private static String getSubdomain(URI uri) { String host = uri.getHost();
```
liga-oz, line 61: what's this private static in javadoc?

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtIssuerValidator.java
```diff
@@ -38,21 +37,6 @@
...
-	 */
-	@Deprecated
-	JwtIssuerValidator(URI url) {
-		assertNotNull(url, "JwtIssuerValidator requires a url.");
-
-		this.domains = Collections.singletonList(url.getHost());
-	}
-
 	/**
 	 * Creates instance of Issuer validation using the domain.
```
liga-oz, line 41: I think this javadoc does not match the method anymore.

java-security/src/test/java/com/sap/cloud/security/token/validation/validators/JwtIssuerValidatorTest.java
```diff
@@ -80,43 +71,38 @@ public void tokenIssuerWithoutHttpSchemeMatchesIdentityProviderUrl() {
...
+		when(token.getClaimAsString(IAS_ISSUER)).thenReturn("https://test.accounts400.ondemand.com");
 		ValidationResult validationResult = cut.validate(token);
 		assertThat(validationResult.isErroneous(), is(true));
 		assertThat(validationResult.getErrorDescription(),
 				startsWith("Issuer validation can not be performed because Jwt token does not contain 'iss' claim."));
 	}
 
 	@Test
-	public void validationFails_whenTokenIssuerIsNull() {
+	public void validationFails_whenIssuerIsNull() {
```
liga-oz, line 105: just a suggestion, maybe use Parameterized Tests here, seems that you have a pattern for tests that could be parameterized like [here](https://github.com/SAP/cloud-security-xsuaa-integration/blob/9713812ddbc8cb353188725b9067a651eb252952/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/XsuaaTokenTest.java#L100)?

liga-oz (changes requested):
just a few unclarities, other than that looks fine 😉

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtIssuerValidator.java
nenaraab: oh my gosh 💯

PR #545, opened 2021-06-21
SpringHttpClient refactoring

liga-oz:
- Refactor bean creation strategy for mTLS 
- deprecate not mTLS supported methods


nenaraab (approved):
💯

PR #546, opened 2021-06-21
enable e2e test for mtls

liga-oz:
- [x] adapt token request to use X.509
- [x] add client secret based Test for tokenFlows 
- [x] refactor base test to create service instances once per test class


nenaraab (commented):
Hi @liga-oz 
thanks for the PR, which also contains the performance improvement to deploy the application once per Test class. I'm pretty sure, that its working - also in multi-threading mode 👍 , but what was the reason to implement it differently than here: https://github.wdf.sap.corp/CPSecurity/cloud-authorization-client-library-java/commit/6b6cb8aca9d4218635fbc88ca921f80c8a2c3a0b#diff-53afa2238379cf02030dfe8117b72fa1 . Unifying it would decrease the maintenance efforts.

I guess the difference is that in the ams python tests, the app get intialized like that
````
@classmethod
def setUpClass(cls):
   cls.cf_app = CFApp(...)
````
The deployment is triggered in the `SampleTest.setUp` instance method, but it returns immediately in case `if(self.deployed_app is not None`.

...and the test class accesses the cf_app via `self.cf_app.deployed_app.clientid` instead of using getters like `self.get_deployed_app().clientid`.

liga-oz:
hi @nenaraab,
in the way how it's setup in ams, I still see that when running a single test class it creates with every test case within the class a new service instance, therefore, i adapted it in a way that it's done only once per test class.

nenaraab (approved):

Issue #549, opened 2021-06-23
Fix HttpHeaders.toString()


nenaraab:
Token was requested for endpoint uri=https://amsserver.authentication.cert.sap.hana.ondemand.com/oauth/token with **headers=HttpHeaders{headers=[com.sap.cloud.security.xsuaa.http.HttpHeader@e6516d15, com.sap.cloud.security.xsuaa.http.HttpHeader@8107e611, com.sap.cloud.security.xsuaa.http.HttpHeader@1a6a91e0]}** and parameters={grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer}

liga-oz:
Merged #555

PR #555, opened 2021-06-30
fix HttpHeaders toString method #549

liga-oz:


token-client/src/test/java/com/sap/cloud/security/xsuaa/http/HttpHeadersTest.java
```diff
@@ -38,9 +38,13 @@ public void equals_differentHeaders_isNotEqual() {
...
 		HttpHeaders headers3 = new HttpHeaders(sameKeyDifferentValue);
 
-		assertThat(headers1).isNotEqualTo(headers2);
-		assertThat(headers1).isNotEqualTo(headers3);
+		assertThat(headers1).isNotEqualTo(headers2).isNotEqualTo(headers3);
 		assertThat(headers2).isNotEqualTo(headers3);
 	}
 
+	@Test
+	public void toStringTest() {
```
nenaraab, line 46: pls just remove "Test" in test case name :-)

nenaraab (approved):

token-client/src/test/java/com/sap/cloud/security/xsuaa/http/HttpHeadersTest.java
liga-oz: @nenaraab it's clashing then with the super toString method

PR #556, opened 2021-07-05
improve logging message for autoconfig

liga-oz:


nenaraab (approved):
Looks much better and thanks for version update 👍

PR #559, opened 2021-07-06
Fix missing java-api dependency in spring-xsuaa

liga-oz:


nenaraab (approved):
Yeah, it comes with token-client lib, but in case we do not check it as part of auto-configuration we have to add it, correct and as it doesn't harm, I'm fine with this additional dependency.

PR #563, opened 2021-07-07
update docu for externally managed private key provisioning

liga-oz:
update readme

### DIFF
--- a/token-client/README.md
+++ b/token-client/README.md
@@ -118,22 +118,32 @@ requestFactory.setHttpClient(HttpClientFactory.create(<OAuth2ServiceConfiguratio
 RestOperations restOperations = new RestTemplate(requestFactory);            
 ```
 
-For X.509 based authentication method using an externally managed certificate, `ClientCertificate` class needs to be instantiated with the external key. For `spring-xsuaa` based applications it can be easily done by providing `xsuaa.key` value as a command-line argument or System environment variable.
-- Executable jar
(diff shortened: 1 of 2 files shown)


nenaraab (commented):
Hi @liga-oz do you know this documentation?
https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config

and don't forget to adapt the `spring-security` project.

nenaraab (approved):
Hi @liga-oz I've incorporated my findings directly. If you are ok we can merge it to master

token-client/README.md
```diff
@@ -168,12 +162,37 @@ To consume the `XsuaaTokenFlows` class, you simply need to `@Autowire` it like t
 private XsuaaTokenFlows xsuaaTokenFlows;
 ```
 
-For X.509 based authentication method using an externally managed certificate, `ClientCertificate` class needs to be instantiated with the external key. For `spring-xsuaa` based applications it can be easily done by overriding these values in `application.yml properties file.
-```yaml
-# For externally managed X.509 certificate
-xsuaa:
-  key: -----BEGIN RSA PRIVATE KEY-----YOUR PRIVATE KEY-----END RSA PRIVATE KEY-----
-```
+For X.509 based authentication method using an externally managed certificate, `ClientCertificate` class needs to be instantiated with the external key and `xsuaaTokenFlows` bean needs to be overwritten using this `ClientCertificate` instance. <br>Alternatively you can provide the certificate key property as command-line argument or programmatically providing default property.
```
liga-oz, line 165: @nenaraab then we don't mention command line argument example? do you think it's not relevant?

nenaraab: exactly. i see no benefit for productive usage :-)

PR #565, opened 2021-07-08
Fix url issue when using X.509

liga-oz:
For TokenBrokerResolver url was not resolved correctly for X.509 authentication method.


nenaraab (approved):
great finding!

PR #571, opened 2021-07-14
Lgtm findings

nenaraab:


spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/XsuaaTokenTest.java
```diff
@@ -181,7 +180,7 @@ public void getZoneIdAsTenantGuid() {
 	}
 
 	@Test
-	public void getSubaccountIdFromSystemAttributes() throws IOException {
+	public void getSubaccountIdFromSystemAttributes() {
```
liga-oz, line 183: `public` access modifier is redundant in Junit5

liga-oz (approved):
Looks good, just if you want, you can get rid of `public` in junit5 test class

PR #607, opened 2021-07-30
k8s support in java-security

liga-oz:


java-security/pom.xml
```diff
@@ -133,6 +133,12 @@
 			<version>${spring.security.version}</version>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
```
nenaraab, line 136: does this replace this one?
````
<dependency>
			<groupId>com.github.stefanbirkner</groupId>
			<artifactId>system-rules</artifactId>
			<version>1.19.0</version>
			<scope>test</scope>
		</dependency>
````

File: java-security/src/main/java/com/sap/cloud/security/config/Environments.java [hunk 2]
```diff
@@ -49,4 +74,8 @@ public static Environment readFromInput(InputStream input) {
 		return CFEnvironment.getInstance(str -> vcapServices.toString(), str -> null);
 	}
 
+	private static boolean isK8sEnv() {
```
nenaraab, line 77: is it required to check that again, otherwise the result should be remembered in a static variable

nenaraab (commented):
we need to improve at least the java doc for method
https://github.com/SAP/cloud-security-xsuaa-integration/blob/535e13b6d714892dc6800fa1f48fced807113821/java-security/src/main/java/com/sap/cloud/security/config/Environments.java#L68

which expects a file in VCAP_SERVICES format

java-security/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -0,0 +1,178 @@
...
+ * supported identity {@link Service} in the Kubernetes Environment by
+ * accessing defaults service secrets paths "/etc/secrets/sapcp/xsuaa" for Xsuaa service or "/etc/secrets/sapcp/ias" for .
+ * IAS service
+ */
+public class K8sEnvironment implements Environment {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(K8sEnvironment.class);
+
+    private static K8sEnvironment instance;
+    private static final String DEFAULT_XSUAA_PATH = "/etc/secrets/sapcp/xsuaa";
```
nenaraab, line 34: should we name it sapcp or sapbtp?

java-security/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -0,0 +1,178 @@
...
+        return getInstance(System::getenv, System::getProperty);
+    }
+
+    private static Map<Service, List<OAuth2ServiceConfiguration>> loadAll() {
+        Map<Service, List<OAuth2ServiceConfiguration>> serviceConfigurations = new HashMap<>(); // NOSONAR
+        List<OAuth2ServiceConfiguration> allXsuaaServices = loadOauth2ServiceConfig(Service.XSUAA);
+        List<OAuth2ServiceConfiguration> allIasServices = loadOauth2ServiceConfig(Service.IAS);
+
+        serviceConfigurations.put(Service.XSUAA, allXsuaaServices);
+        serviceConfigurations.put(Service.IAS, allIasServices);
```
nenaraab, line 70: allIasServices -> iasService

nenaraab (changes requested):
Hi @liga-oz 
could you please add K8SEnvironmentTest here?
Thanks!
Nena

samples/java-security-usage/k8s/deployment.yml
```diff
@@ -0,0 +1,64 @@
...
+          volumeMounts:
+            - name: xsuaa
+              mountPath: "/etc/secrets/sapcp/xsuaa/xsuaa-instance"
+              readOnly: true
+            - name: xsuaa2
+              mountPath: "/etc/secrets/sapcp/xsuaa/xsuaa-instance-1"
+              readOnly: true
+
+          envFrom:
+            - secretRef:
```
nenaraab, line 39: what is meant with that optional secretRef?

nenaraab:
i will skip this PR and check PR #610 instead, it seems, that it provides the missing test case and the multiple xsuaa bindings are supported there...

java-security/pom.xml
liga-oz: it's for environment variable testing support in jupiter, as some tests still use junit4 can't replace it completely plus I run into issue when jupiter test didn't work with environment variable and i needed to switch back to junit4 to get it working with surefire plugin

java-security/src/test/java/com/sap/cloud/security/config/EnvironmentsTest.java
```diff
@@ -26,20 +30,36 @@ public EnvironmentsTest() throws IOException {
...
 
 		assertThat(firstEnvironment).isSameAs(secondEnvironment);
 	}
 
 	@Test
-	public void getCurrent_returnsCorrectEnvironment() {
-		// TODO 29.11.19 c5295400: extend test when more than one environment is
-		// supported
+	public void getCurrent_returnsOnlySingleK8sInstance(EnvironmentVariables environmentVariables) {
+		environmentVariables.set("KUBERNETES_SERVICE_HOST", "0.0.0.0");
```
nenaraab, line 42: please use an constant

samples/java-security-usage/k8s/deployment.yml
liga-oz: this is removed in later PR

java-security/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
liga-oz: @nenaraab i guess this docu is then outdated https://wiki.wdf.sap.corp/wiki/display/PFS/Guidelines+for+SAP+CP+Service+Libraries 
to me it also makes sense to rename it to sapbtp to be inline with the new name

liga-oz:
> Hi @liga-oz
> could you please add K8SEnvironmentTest here?
> Thanks!
> Nena

test is available in #610

java-security/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
liga-oz: @nenaraab I would propose to rename it in separate PR as it affects all later created PRs

nenaraab: sure 👍

nenaraab (approved):

PR #614, opened 2021-08-06
k8s spring-security-basic-auth sample enhancement

liga-oz:


samples/spring-security-basic-auth/k8s/deployment.yml
```diff
@@ -0,0 +1,168 @@
...
+          secret:
+              secretName: service-manager-binding
+---
+apiVersion: v1
+kind: Service
+metadata:
+  creationTimestamp: null
+  labels:
+    app: spring-security-basic-auth
+  name: spring-security-basic-auth
```
nenaraab, line 134: intendation issue

nenaraab (changes requested):
looks very good - beside of the intendation

samples/spring-security-basic-auth/k8s/deployment.yml
liga-oz: that's meant like that, label with `app` as a key and app name as value and then the service `name`, i guess it's confusing because app and service has the same names.

nenaraab (approved):

PR #617, opened 2021-08-10
k8s spring-security and spring-security-hybrid-usage sample enhancement

liga-oz:


nenaraab (approved):
Hi @liga-oz 

great! I was able to deploy and test the application (but without authorization, as it was not possible to assign the RoleCollection to the client). I guess we need to elaborate that and check whether we can adapt our environment to a custom ias to test the same with user token.
The other comments I have already incorporated.

Another thing i was not sure, whether there is a duplicate:
- there are two api rules: spring-security-hybrid-api and spring-hybrid, which seems to return the same results... pls check.

samples/spring-security-hybrid-usage/k8s/deployment.yml
```diff
@@ -0,0 +1,214 @@
...
+        description: View Data
+        scope-references:
+          - "$XSAPPNAME.Read"
+    role-collections:
+      - name: XSUAA-Viewer
+        description: Viewer (read)
+        role-template-references:
+          - "$XSAPPNAME.Viewer"
+---
+apiVersion: servicecatalog.k8s.io/v1beta1
```
nenaraab, line 29: according to Jakob there is a successor of this service catalog, named BTP service operator:
https://github.com/SAP/sap-btp-service-operator/blob/main/README.md

Even this is still in alpha state (release planned for Q4) I think we should think about supporting only this but not catalog. With this plugin not each property is stored in its own file, but there is one file with all properties, such as clientid, certificate,...

PR #626, opened 2021-08-17
Expose token_type in OAuth2TokenResponse

### LINKED ISSUE #606: Add token_type in OAuth2TokenResponse
Hi Colleagues,

We are trying to use the XSUAA Token Client and Token Flow API for token retrieval.

However, there is one little problem.

If we look at OAuth2TokenResponse we can see that it contains the access token and its expiration time, but it doesn't contain the token type.

Could you please add it to the OAuth2TokenResponse so that there won't be need to use a hard-coded "Bearer " string in our code?

Best regards,
Tsvetomir Lyubenov

liga-oz:
Fixes #606

### DIFF
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenResponse.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenResponse.java
@@ -24,12 +24,14 @@ public class OAuth2TokenResponse {
 		this(accessToken, expiredInSeconds, refreshToken, TOKEN_TYPE);
 	}
 
-	public OAuth2TokenResponse(@Nullable String accessToken, long expiredInSeconds, @Nullable String refreshToken, String tokenType) {
+	public OAuth2TokenResponse(@Nullable String accessToken, long expiredInSeconds, @Nullable String refreshToken,
+			String tokenType) {
 		this.accessToken = accessToken;
 		this.expiredTimeMillis = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(expiredInSeconds);
 		this.refreshToken = refreshToken;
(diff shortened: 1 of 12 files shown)


token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenResponse.java
```diff
@@ -24,6 +25,10 @@ public OAuth2TokenResponse(@Nullable String accessToken, long expiredInSeconds,
 		this.refreshToken = refreshToken;
 	}
 
+	public OAuth2TokenResponse(@Nullable String accessToken, long expiredInSeconds, @Nullable String refreshToken, String tokenType) {
```
nenaraab, line 28: i would turn it around, existing constructor calls new one, tokenType can be defaulted to "bearer" (see also [here](https://docs.cloudfoundry.org/api/uaa/version/4.30.0/index.html#jwt-bearer-token-grant
)).

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/XsuaaOAuth2TokenServiceClientCredentialsTest.java
```diff
@@ -5,22 +5,6 @@
  */
 package com.sap.cloud.security.xsuaa.client;
 
-import static com.sap.cloud.security.xsuaa.client.OAuth2TokenServiceConstants.ACCESS_TOKEN;
```
nenaraab, line 8: have you applied formatter?

nenaraab (changes requested):
only minor change requested in constructor. Maybe this simplifies the test setup?

nenaraab (approved):

PR #627, opened 2021-08-17
XsuaaServiceConfigurationDefault supports access to other VCAP_SERVICES credentials properties

### LINKED ISSUE #601: XsuaaServiceConfigurationDefault Property Extension
Hi colleagues,

Our team uses the `XsuaaServiceConfigurationDefault` configuration from `spring-security-xsuaa` and we have the use case to get the `tenantid` and `apiurl` properties of an XSUAA service instance.

Could you extend the `XsuaaServiceConfigurationDefault` configuration with getters for the mentioned properties? 
As I can see, the `XsuaaServicePropertySourceFactory` can be extended, so they can be mapped to the configuration.

Thanks and best regards,
Momchil

nenaraab:
- cleanup "identityzoneid" as it was never exposed
- `XsuaaServiceConfigurationDefault.hasProperty("apiurl")` returns true if VCAP_SERVICES-xsuaa-credentials contains attribute "apiurl"
- `XsuaaServiceConfigurationDefault.getProperty("apiurl")` returns value from VCAP_SERVICES-xsuaa-credentials-apiurl or null, if attribute does not exist.

closes #601

### DIFF
--- a/java-security/README.md
+++ b/java-security/README.md
@@ -89,7 +89,7 @@ OAuth2ServiceConfiguration serviceConfig = OAuth2ServiceConfigurationBuilder.for
       .withClientSecret("oauth-client-secret")
       .build();
 ```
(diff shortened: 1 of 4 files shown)


liga-oz (approved):
👍🏻

nenaraab:
Hi @m-ignatov you can precheck that solution... will be released in the next hours / days.

PR #632, opened 2021-08-19
Extend XSUAA Service Configuration properties

m-ignatov:
- Add the following properties to the XSUAA credentials support by the library: apiurl, tenantid, subaccountid
follow up of #601


nenaraab:
Hi @m-ignatov 

thanks for your contribution!

I've provided you a lightweight solution on how to access these additional fields in version 2.10.4.
In case you have already seen that, what do you miss?

When I add your enhancements here, I need to offer that as well in `java-security` and `spring-security` project, and the interface is exactly the same with `hasProperty` and `getProperty` methods which allows access to additional properties that are XSUAA or IAS specific. 

Furthermore, as Torsten Luh mentioned exposing the tenant_id might be confusing as normally developers should derive it from the token itself (`getZoneId()`).

Kind regards,
Nena

m-ignatov:
Hi @nenaraab 

Thank you for the proposed solution and fast reply.

However, as a consumer of the XSUAA library (version 2.10.4), we now have to maintain a list of the XSUAA VCAP properties names, which does not seem to me as a consumer responsibility, in terms of separation of concerns and isolation.

Moreover, as I can see in the library, there is already a list of XSUAA credential properties you support, so why not add these one as well, and their corresponding getters/ setters?

We would like to retrieve these XSUAA properties in an unified and simple manner, like for example:
- `xsuaaServiceConfigurationDefault.getClientId();`
- `xsuaaServiceConfigurationDefault.getClientSecret();`
- ...
- `xsuaaServiceConfigurationDefault.getApiUrl();` 
instead of `xsuaaServiceConfigurationDefault.getProperty("apiurl");` 


It would be great if you can consider this approach, and if needed, I would also make the needed changes to this PR.

Kind regards,
Momchil

nenaraab:
Hi @m-ignatov 

yes, we can consider to introduce some constants, but thats all I want to provide here for the reasons i mentioned above.

Kind regards,
Nena

m-ignatov:
Yes, this would be helpful.

I saw there is a class `CFConstants` and suppose there will be added the new constants?
I can update this PR with the added constants only.

Kind regards,
Momchil

m-ignatov:
Hi @nenaraab 

I have made the changes we discussed.

Please, let me know if you have any comments on the PR.

Kind regards,
Momchil

nenaraab (approved):
👍 very good. Thanks for the contribution!

PR #636, opened 2021-08-20
introduce new env project and deprecated XsuaaServicesParser (spring-xsuaa)

nenaraab:
This is a preparation for K8S when spring-xsuaa uses Environments.getCurrent() for parsing xsuaa service binding.


env/src/main/java/com/sap/cloud/security/config/Environments.java
```diff
@@ -5,6 +5,7 @@
  */
 package com.sap.cloud.security.config;
 
+
 import com.sap.cloud.security.config.cf.CFEnvironment;
```
liga-oz, line 9: Did you apply formatter?

env/src/test/java/com/sap/cloud/security/config/EnvironmentsCFTest.java
```diff
@@ -16,32 +16,30 @@
...
 		Environment secondEnvironment = Environments.getCurrent();
 
 		assertThat(firstEnvironment).isSameAs(secondEnvironment);
 	}
 
 	@Test
-	public void getCurrent_returnsCorrectEnvironment() {
-		// TODO 29.11.19 c5295400: extend test when more than one environment is
-		// supported
+	void getCurrent_returnsCf() {
```
liga-oz, line 37: Cf or CF abbreviation should be consistent throughout all tests and for this test I think it suit even better to rename it to getType_returnsCF

env/src/test/java/com/sap/cloud/security/config/json/DefaultJsonObjectTest.java
```diff
@@ -3,8 +3,11 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-package com.sap.cloud.security.json;
+package com.sap.cloud.security.config.json;
```
liga-oz, line 6: this is not the right package, you need to move it outside config package

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaAudienceValidator.java
```diff
@@ -32,7 +31,7 @@
  */
 public class XsuaaAudienceValidator implements OAuth2TokenValidator<Jwt> {
 	private Map<String, String> appIdClientIdMap = new HashMap<>();
-	private final Logger logger = LoggerFactory.getLogger(XsuaaServicesParser.class);
+	private final Logger logger = LoggerFactory.getLogger(XsuaaAudienceValidator.class);
```
liga-oz, line 34: great finding 👍🏻

liga-oz (changes requested):
The json package is located in the config package `com.sap.cloud.security.config.json` which is different path then previously, it needs to be in `com.sap.cloud.security.json`, other than that some minor things to consider.

env/src/main/java/com/sap/cloud/security/config/Environments.java
nenaraab: yes, finally I've applied the formatting rules, should be gone 👍

env/src/test/java/com/sap/cloud/security/config/json/DefaultJsonObjectTest.java
nenaraab: great finding. should be corrected now.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationDefaultTest.java
```diff
@@ -0,0 +1,41 @@
+package com.sap.cloud.security.xsuaa;
+
+import com.sap.cloud.security.config.cf.CFConstants;
+import org.junit.Before;
```
liga-oz, line 4: Was there a specific reason why you chose junit4 instead of junit5? If not I would suggest to write all new tests in Junit5.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServicePropertySourceFactoryMultipleBindingsTest.java
```diff
@@ -0,0 +1,48 @@
+package com.sap.cloud.security.xsuaa;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
```
liga-oz, line 4: same as above junit4 instead of junit5

liga-oz (changes requested):
Just double check if `reuseFork` is required in spring-xsuaa project. 
Should junit5 be used in all new tests instead of junit4?

spring-xsuaa/pom.xml
```diff
@@ -194,6 +206,9 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
+				<configuration>
```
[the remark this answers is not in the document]
nenaraab: Yeah, i've understood, that EnvironmentVariables Rule didn't worked with JUnit5?

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServicePropertySourceFactoryMultipleBindingsTest.java
nenaraab: Here I just copied the test cases from XsuaaServicesParserTest, which were written in Junit 4

spring-xsuaa/pom.xml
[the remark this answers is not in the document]
nenaraab: Unfortunately I need that, otherwise XsuaaServiceConfigurationDefaultTest.getProperty:28 fails with NullPointer

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationDefaultTest.java
nenaraab: Here I just copied the test cases from XsuaaServicesParserTest, which were written in Junit 4. Furthermore i've understood, that EnvironmentVariables Rule didn't worked with JUnit5?

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServicePropertySourceFactoryMultipleBindingsTest.java
liga-oz: oki 👍🏻

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationDefaultTest.java
liga-oz: the `Rule` itself doesn't work with junit5 it needs to be migrated to `SystemStubs` `EnvironmentVariables`, which is another library, but  Junit5 `EnvironmentVariables` works overall with Junit5 the problem was the test isolation for different environment variable settings, which requires then `reuseForks` to be set to false in surefire plugin.

liga-oz (approved):
all good 👍🏻

PR #640, opened 2021-08-24
SECAUTH-1483

nenaraab:
### In `spring-xsuaa` 
**Before**
```java
Converter<Jwt, AbstractAuthenticationToken> getJwtAuthenticationConverter() {
		TokenAuthenticationConverter converter = new TokenAuthenticationConverter(xsuaaServiceConfiguration);
		converter.setLocalScopeAsAuthorities(true);
		return converter;
}
```
**New**
```java
Converter<Jwt, AbstractAuthenticationToken> getJwtAuthenticationConverter() {
          return new TokenAuthenticationConverter(new MyCustomAuthoritiesExtractor());
}

private static class MyCustomAuthoritiesExtractor implements AuthoritiesExtractor {

		@Override
		public Collection<GrantedAuthority> getAuthorities(XsuaaToken token) {
			String appId = "";
			if(token.getClientId().startsWith("sb-")) {
				appId = token.getClientId().replaceFirst("sb-", "");
			}
			AuthoritiesExtractor authoritiesExtractor = new LocalAuthoritiesExtractor(appId);
			return authoritiesExtractor.getAuthorities(token);
		}

	}
```
### In `spring-security` 
You can subclass `XsuaaTokenAuthorizationConverter` and overwrite `localScopeAuthorities(Jwt jwt, Collection<String> scopes)`. An example is not yet available.

### DIFF
--- a/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/XsuaaTokenAuthorizationConverter.java
+++ b/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/XsuaaTokenAuthorizationConverter.java
@@ -41,14 +41,18 @@ public class XsuaaTokenAuthorizationConverter implements Converter<Jwt, Abstract
 	}
 
 	protected Collection<GrantedAuthority> localScopeAuthorities(Jwt jwt) {
-		Collection<GrantedAuthority> localScopeAuthorities = new ArrayList<>();
 		Collection<String> scopes = jwt.getClaimAsStringList(TokenClaims.XSUAA.SCOPES);
 		if (scopes == null) {
 			return Collections.emptySet();
 		}
+		return localScopeAuthorities(jwt, scopes);
+	}
+
+	protected Collection<GrantedAuthority> localScopeAuthorities(Jwt jwt, Collection<String> scopes) {
+		Collection<GrantedAuthority> localScopeAuthorities = new ArrayList<>();
 		for (String scope : scopes) {
 			if (scope.startsWith(appId + ".")) {
-				localScopeAuthorities.add(new SimpleGrantedAuthority(scope.replaceFirst(appId + ".", "")));
+				localScopeAuthorities.add(new SimpleGrantedAuthority(scope.substring(appId.length() + 1)));
 			}
 		}
 		return localScopeAuthorities;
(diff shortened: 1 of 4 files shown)


spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/XsuaaTokenAuthorizationConverter.java
```diff
@@ -41,14 +41,18 @@ public AbstractAuthenticationToken convert(Jwt jwt) {
...
 	protected Collection<GrantedAuthority> localScopeAuthorities(Jwt jwt) {
-		Collection<GrantedAuthority> localScopeAuthorities = new ArrayList<>();
 		Collection<String> scopes = jwt.getClaimAsStringList(TokenClaims.XSUAA.SCOPES);
 		if (scopes == null) {
 			return Collections.emptySet();
 		}
+		return localScopeAuthorities(jwt, scopes);
+	}
+
+	protected Collection<GrantedAuthority> localScopeAuthorities(Jwt jwt, Collection<String> scopes) {
```
liga-oz, line 51: the jwt seems to be not used in the method, is it required to have it in the argument list?

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/TokenAuthenticationConverterTest.java
```diff
@@ -141,4 +156,18 @@ private static String getSidForAttributeValue(String attributeName, String attri
...
+		@Override
+		public Collection<GrantedAuthority> getAuthorities(XsuaaToken token) {
+			String appId = "";
+			if(token.getClientId().startsWith("sb-")) {
+				appId = token.getClientId().replaceFirst("sb-", "");
+			}
+			AuthoritiesExtractor authoritiesExtractor = new LocalAuthoritiesExtractor(appId);
+			return authoritiesExtractor.getAuthorities(token);
+		}
+
```
liga-oz, line 171: have you applied formatter?

liga-oz (changes requested):
All looking good just couple of points to review 😉

spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/XsuaaTokenAuthorizationConverter.java
nenaraab: good finding! 
But in this case this protected method is foreseen as "enhancement spot" and app developers may need access to other token claims.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/token/TokenAuthenticationConverterTest.java
nenaraab: done now :-)

liga-oz (approved):
👍🏻

PR #645, opened 2021-08-27
SSLContextFactory.createKeyStore(ClientIdentity)

### LINKED ISSUE #621: [token-client] extend SSLContextFactory with KeyStore getter method
For a case when application provides custom implementation for `HttpClientFactory` interface, it would be helpful to have access to `KeyStore` object to reduce code duplications for certificate and key parsing when initialising custom `SSLContext`.

Proposal to add public method `getKeyStore` to [SSLContextFactory](https://github.com/SAP/cloud-security-xsuaa-integration/blob/master/token-client/src/main/java/com/sap/cloud/security/mtls/SSLContextFactory.java) which would resolve accordingly the `certificate` and `key` attributes from `ClientIdentity` class and initialize the `KeyStore`.
```java
public KeyStore getKeyStore(ClientIdentity clientIdentity){...}
```

nenaraab:
fixes #621  (cloud sdk HttpDestinationProperties)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 All notable changes to this project will be documented in this file.
 
 ## 2.10.5
-- [token-client] 
+- [token-client]
+  - new method `SSLContextFactory.createKeyStore(ClientIdentity)`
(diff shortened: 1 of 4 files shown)


liga-oz (approved):
👍🏻

Issue #647, opened 2021-09-02
Adapt ias samples to use X.509 instead of client secrets


nenaraab:
Example
````
  "identity": [
    {
      "credentials": {
        "certificate": "-----BEGIN CERTIFICATE-----\nMIIGIDC...\n-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----\nMIIGaDC...\n-----END CERTIFICATE-----",
        "certificate_expires_at": "2021-10-02T09:27:59Z",
        "clientid": "773a85c6-5e69-49a3-87dc-3dbc5274bd33",
        "domains": [
          "accounts400.ondemand.com"
        ],
        "key": "-----BEGIN RSA PRIVATE KEY-----\nMI...==\n-----END RSA PRIVATE KEY-----",
        "url": "https://mytenant.accounts400.ondemand.com",
        "zone_uuid": "..."
      },
      "instance_name": "authn",
      "label": "identity",
      "name": "authn",
      "plan": "application"
    }
  ]
}
````
So there exists no `credential-type` as you have it with xsuaa.

nenaraab:
Hi @liga-oz can you pre-analyze if it is an issue, when `credential-type` is not provided in context of ias (I've seen it in the python tests...). Do you expect this to be existent in `java-security` and/or `spring-security`?

nenaraab:
@liga-oz good idea!

nenaraab:
reopen, as samples needs to be adapted...

PR #651, opened 2021-09-06
K8s extension

liga-oz:
follow up PR of this one #630


env/pom.xml
```diff
@@ -24,18 +24,15 @@
...
         </dependency>
         <dependency>
-            <groupId>commons-io</groupId>
-            <artifactId>commons-io</artifactId>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
+            <groupId>com.sap.cloud.security.xsuaa</groupId>
```
nenaraab, line 27: pls add a comment that token-client and http-client are only required, in case of multiple xsuaa service bindings.

java-api/src/main/java/com/sap/cloud/security/config/k8s/K8sConstants.java
```diff
@@ -0,0 +1,48 @@
...
+import com.sap.cloud.security.config.cf.CFConstants;
+
+/**
+ * Constants that simplifies access to service configuration properties in the
+ * Kubernetes environment.
+ */
+public class K8sConstants {
+	public static final String KUBERNETES_SERVICE_HOST = "KUBERNETES_SERVICE_HOST";
+	static final String XSUAA_CONFIG_PATH_DEFAULT = "/etc/secrets/sapbtp/xsuaa";
+	static final String IAS_CONFIG_PATH_DEFAULT = "/etc/secrets/sapbtp/identity";
```
nenaraab, line 17: maybe we shall rename it IDENTITY_CONFIG_PATH_DEFAULT

samples/java-security-usage/README.md
```diff
@@ -1,7 +1,83 @@
...
+- Get an access token via `curl`. You can get the information to fill the placeholders from the service binding secret:
+```shell script
+kubectl get secret "xsuaa-service-binding" -o go-template='{{range $k,$v := .data}}{{"### "}}{{$k}}{{"\n"}}{{$v|base64decode}}{{"\n\n"}}{{end}}' -n <YOUR NAMESPACE>
+```
+
+```
+curl -X POST \
+  <<url>>/oauth/token \
+  -H 'Content-Type: application/x-www-form-urlencoded' \
+  -d 'client_id=<<clientid>>&client_secret=<<clientsecret>>&grant_type=password&username=<<your username>>&password=<<your password>>'
```
nenaraab, line 58: does this fit to X.509 client credential setup?

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/TestController.java
```diff
@@ -32,4 +32,9 @@
 		result.put("scopes", String.valueOf(token.getScopes()));
 		return result;
 	}
+
+	@GetMapping("/health")
```
nenaraab, line 36: was this required by K8S?

samples/spring-security-hybrid-usage/README.md
```diff
@@ -2,10 +2,76 @@
...
+
+## Access the application
+- get an IAS oidc token via ``password`` grant token flow. For that call the ``/oauth2/token`` endpoint of your identity service. You can get the ``url`` and the ``clientid`` and ``clientsecret`` for the Basic Authorization header from the `ias-service-binding` secret.
+    ```shell script
+      kubectl get secret "ias-service-binding" -o go-template='{{range $k,$v := .data}}{{"### "}}{{$k}}{{"\n"}}{{$v|base64decode}}{{"\n\n"}}{{end}}' -n <YOUR NAMESPACE>
+    ```
+- get a XSUAA access token via ``client`` token flow. For that, call the ``/oauth/token`` endpoint of your xsuaa service. <br>You can get the ``url``, the ``client_id``, the ``client_secret`` for the request from the `xsuaa-service-binding` secret.
+    ```shell script
+      kubectl get secret "xsuaa-service-binding" -o go-template='{{range $k,$v := .data}}{{"### "}}{{$k}}{{"\n"}}{{$v|base64decode}}{{"\n\n"}}{{end}}' -n <YOUR NAMESPACE>
+    ```
```
nenaraab, line 58: i think something is missing here, the way on how to get an token...?

samples/spring-security-hybrid-usage/k8s/deployment.yml
```diff
@@ -0,0 +1,203 @@
...
+  name: ias-authn
+spec:
+  clusterServiceClassExternalName: identity
+  clusterServicePlanExternalName: application
+---
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceBinding
+metadata:
+  name: ias-service-binding
+spec:
```
nenaraab, line 66: in order to leverage X.509 we have to parameterize the IAS service binding:
parameters: { "credential-type": "X509_GENERATED" }

samples/spring-security-hybrid-usage/k8s/deployment.yml
```diff
@@ -0,0 +1,203 @@
...
+      name: ias_
+  serviceBindingRef:
+    name: ias-service-binding
+  usedBy:
+    kind: deployment
+    name: spring-security-hybrid-usage
+---
+# Service manager instance
+apiVersion: servicecatalog.k8s.io/v1beta1
+kind: ServiceInstance
```
nenaraab, line 86: service manager is not required here

nenaraab (changes requested):
Overall very nice. I've added some minor remarks especially for the samples.

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/TestController.java
liga-oz: nope, but I really like to have a health endpoint, it helps to analyze issue in case app doesn't work properly, mainly for new feature development and testing

samples/spring-security-hybrid-usage/k8s/deployment.yml
liga-oz: shouldn't we leave at least one sample with sm definition and maybe even multiple xsuaa bindings?

samples/spring-security-hybrid-usage/README.md
liga-oz: do you suggest adding a curl command here for the post request to token endpoint?

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/TestController.java
nenaraab: all right :-)

samples/spring-security-hybrid-usage/k8s/deployment.yml
nenaraab: mhhh... for e2e testing it would be good to have it, and hybrid sample might be the one to choose, but on the other side, i would like to keep them as simple as possible...

samples/java-security-usage/README.md
nenaraab: But wouldn't you fetch an access token with certificate and clientsecret isn't provided with secrets at all?

liga-oz: no and i also noticed that this sample is not configured to use X.509 also in CF, should we then adapt both envs to use mTLS?

samples/spring-security-hybrid-usage/k8s/deployment.yml
liga-oz: I think we need to sacrifice one of the samples for now until, we don't need to use the sm anymore.

samples/java-security-usage/README.md
nenaraab: that's right, in cf version its alos not yet adapted... we can also keep it here as it is for now...

samples/spring-security-hybrid-usage/README.md
nenaraab: see also here:
https://github.com/SAP/cloud-security-xsuaa-integration/tree/main/samples/spring-security-hybrid-usage#access-the-application
I think it would be good to clean up our repo and to put the different ways on how to get an access token into a separate README, so that we can link to them as part of our sample READMEs.

nenaraab: opened a new issue #652

samples/java-security-usage/README.md
liga-oz: good finding right on time 💯

samples/spring-security-hybrid-usage/k8s/deployment.yml
liga-oz: then i guess CF samples also have to be adapted, or we better  include it in the PR for #647

nenaraab: pls not as part of this PR as the CF sample adaption would also require adaptions in the python scripts

nenaraab:
**This enhancement is available with pre-release 2.11.0-M1** 🥇

nenaraab (approved):

PR #661, opened 2021-09-14
[CAP] OAuth2ServiceConfiguration simplification for implementations

### LINKED ISSUE #658: [CAP] OAuth2ServiceConfiguration improvements
.... in order to simplify creation via `OAuth2ServiceConfigurationBuilder` or creation from scratch

nenaraab:
closes #658

### DIFF
--- a/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationCustom.java
+++ b/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationCustom.java
@@ -13,7 +13,6 @@ import com.sap.cloud.security.config.CredentialType;
 import javax.annotation.Nullable;
 import java.net.URI;
 
-
 public class XsuaaServiceConfigurationCustom implements XsuaaServiceConfiguration {
 
 	private final XsuaaCredentials credentials;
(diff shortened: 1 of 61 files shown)


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaServiceConfigurationCustom.java
```diff
@@ -72,4 +63,17 @@ public CredentialType getCredentialType() {
...
 	}
 
+	@Override
+	public String getProperty(String name) {
+		return null;
+	}
+
+	@Override
+	public ClientIdentity getClientIdentity() {
+		ClientIdentity identity = new ClientCredentials(getClientId(), getClientSecret());
```
liga-oz, line 73: I would reverse the logic -> first create ClientCertificate and check if it is valid and if not then create ClientCredentials as the X.509 becomes soon the new default auth type

liga-oz (changes requested):
I really like it, well done 💯 only one thing to consider the sequence of ClientIdentity instance creation.

nenaraab:
> I really like it, well done 💯 only one thing to consider the sequence of ClientIdentity instance creation.

Hi @liga-oz 
thanks for the hint, you're right, but in the default implementation i'm unable to change that (as getProperty is not implemented everywhere, like in spring-xsuaa) and in XsuaaServiceConfigurationCustom i kept the order from default implementation for simplicity. But we can revisit the sama as soon as ias provides credential-type as part of the binding.

liga-oz (approved):
I really like it, well done 💯 only one thing to consider the sequence of ClientIdentity instance creation.

PR #662, opened 2021-09-14
Introduce extension point for CAP: EnvironmentLoader

nenaraab:


env/src/main/java/com/sap/cloud/security/config/Environments.java
```diff
@@ -6,20 +6,29 @@
...
 /**
  * Central entry point to access the current SAP Cloud Platform
  * {@link Environment}.
  */
 public class Environments {
-
-	private static final Environment cfEnvironment = CFEnvironment.getInstance(); // singleton
+	static List<EnvironmentLoader> environments = new ArrayList() {
+		{
+			ServiceLoader.load(EnvironmentLoader.class).forEach(this::add);
```
BraunMatthias, line 25: Not clear how clients can overrule the default loader as all matching loaders will be added to environments, but only the one with index 0 is taken as current.
What about replacing the ServiceLoader API and just provide a static setter which writes a single reference (default to DefaultEnvironmentLoader) instead of a list? This also helps in a Spring context in which ServiceLoader implementations are hard to combine with initialization of Spring beans.

nenaraab: Hi @BraunMatthias  thanks for the feedback! We will revisit your first question soon. 

What do you mean with `This also helps in a Spring context in which ServiceLoader implementations are hard to combine with initialization of Spring beans`?
In our AMS client lib, we heavily combine Service loader and spring-beans....

env/src/main/java/com/sap/cloud/security/config/EnvironmentLoader.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * SPDX-FileCopyrightText: 2018-2021 SAP SE or an SAP affiliate company and Cloud Security Client Java contributors
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.sap.cloud.security.config;
+
+public interface EnvironmentLoader {
```
liga-oz, line 8: java doc missing

env/src/main/java/com/sap/cloud/security/config/Environments.java
BraunMatthias: My specific EnvironmentLoader needs some of our Spring beans as prerequisite (as input parameter) which also puts some constraints on creation order etc. Ideally, the EnvironmentLoader  should be bean itself that dependes on other beans in my setup. On the other hand, it seems there is no control about service loader object creation and input parameters (the creation is triggered when the Environments class is loaded, no input parameters).
Can you point me to an example how to deal with objects loaded via ServiceLoader that rely on Spring bean capabilities?

liga-oz (changes requested):
please provide missing java docs for the new classes/interface

nenaraab:
Close this for now, as CAP needs to overwrite environment as part of Spring EnvironmentPostProcessor, see here: https://www.baeldung.com/spring-boot-environmentpostprocessor

PR #672, opened 2021-09-20
fix Internal Server error caused by getClientIdentity method 

liga-oz:
As Xsuaa still delivers the clientSecret for X.509 credential type, logic to resolve the clientIdentity fails and results to Internal Server Error.
As a fix I propose to swap around creation of `ClientIdentity` implementation classes -> Try to resolve `ClientCertificate` first, if it's not valid only then `ClientCredentials`.
Regarding the getProperty method that is by default not supported in `XsuaaServiceConfiguration` interface all of our implementations now overrides this method. 
In case of custom implementations also the way around approach in getClientIdentity method that resolves `ClientCredentials` first would result also in `UnsupportedOperationException` exception once the clientsecret was removed from Xsuaa credentials in case of X.509.

### DIFF
--- a/spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
+++ b/spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
@@ -46,9 +46,12 @@ class OAuth2ServiceConfigurationPropertiesTest {
 	}
 
 	@Test
-	void setGetCertificateAndKey() {
+	void setGetCertificateAndKeyIAS() {
 		cutIas.setKey(ANY_VALUE);
 		cutIas.setCertificate(ANY_VALUE);
+		cutIas.setClientId(ANY_VALUE);
+		cutIas.setClientSecret(ANY_VALUE); // to make sure that getClientIdentity uses ClientCertificate impl as default
+											// when possible
 		assertEquals(ANY_VALUE, cutIas.getClientIdentity().getKey());
 		assertEquals(ANY_VALUE, cutIas.getClientIdentity().getCertificate());
 		assertTrue(cutIas.getClientIdentity().isCertificateBased());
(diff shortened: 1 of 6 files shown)


spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
```diff
@@ -58,8 +61,38 @@ void setGetCertificateAndKey() {
 		assertEquals(ANY_VALUE, cutIas.getProperty(CERTIFICATE));
 	}
 
+	@Test
+	void setGetCertificateAndKeyXSUAA() {
```
nenaraab, line 65: whats the difference between setGetCertificateAndKeyIAS and setGetCertificateAndKeyXsuaa?
furthermore, i would expect cut.getClientSecret() is SET

spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
```diff
@@ -58,8 +61,38 @@ void setGetCertificateAndKey() {
...
+		assertTrue(cutXsuaa.getClientIdentity().isCertificateBased());
+	}
+
+	@Test
+	void getClientIdentityResolvesToClientCredentials() {
+		cutIas.setClientId(ANY_VALUE);
+		cutIas.setClientSecret(ANY_VALUE);
+		assertFalse(cutIas.getClientIdentity().isCertificateBased());
+
+		cutXsuaa.setClientId(ANY_VALUE);
```
nenaraab, line 86: whats the difference between cutXsuaa and cutIas?

spring-xsuaa-it/src/test/java/com/sap/cloud/security/xsuaa/extractor/XsuaaServiceConfigurationDummy.java
```diff
@@ -46,4 +46,9 @@ public String getVerificationKey() {
 		return verificationKey;
 	}
 
+	@Override
+	public String getProperty(String name) {
+		return null;
```
nenaraab, line 51: pls add a TODO that we revisit that as soon as IAS provides credential-type.

spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/DummyXsuaaServiceConfiguration.java
```diff
@@ -55,4 +55,9 @@ public String getVerificationKey() {
 	public CredentialType getCredentialType() {
 		return null;
 	}
+
+	@Override
+	public String getProperty(String name) {
```
nenaraab, line 60: pls add a TODO that we revisit that as soon as IAS provides credential-type.

nenaraab (changes requested):
Hi @liga-oz 
great finding! I personally wasn't aware of that secret is part of the X509 binding... and there was no test case for that - so, thanks for the fix!

spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
liga-oz: Internally the only difference is the service field IAS or Xsuaa, but I think to represent both service differences, I would actually change it slightly: for setGetCertificateAndKeyIAS I wouldnt set the clientsecret, only for setGetCertificateAndKeyXSUAA, but then again it doesn't seem to be the right place to test it. I'll combine the both then.

spring-security/src/test/java/com/sap/cloud/security/spring/config/OAuth2ServiceConfigurationPropertiesTest.java
liga-oz: good point, the only difference is the service field

liga-oz: leaving it in as discussed

nenaraab (approved):

PR #677, opened 2021-09-28
[java-security-test] replace wiremock-jre8-standalone with wiremock-jre8

nenaraab:
a stakeholder reported:
image


liga-oz (approved):
nice 👍🏻

PR #704, opened 2021-11-04
Prioritize SPI loading of custom implementations

### LINKED ISSUE #668: Make SPI loading bulletproof
1. Decision:
A) Default implementation is NOT loaded via SPI but . Advantage: we only need to check whether there is 0 or 1 occurrence.
B) as is, default implementation is also registered as SPI. Advantage: in case of `TokenFactory`, the default implementation `HybridTokenFactory` can't be instantiated within `TokenFactory` interface as its not part of java-api. Consequently in case of more than one SP we need to filter our default implementation to find the customers implementation. There should be never more than 2 SPs (incl. our default implementation)

=> **Therefore for consistency purposes we decide for **Option B**, as we can implement it for both cases `TokenFactory` and `HttpClientFactory` the same.**

2. Decision:
A) raise exception, this would mean, the application does not come up and forces the app developer to analyse the configuration issue. May lead to consulting efforts as it is an incompatible change.
B) log error, app developer may oversee the error too late and face behavioural issues because of misconfiguration. An advantage might be in a stacked context, that an overwrite might be possible (??? unable to find documentation about it).

=> **Option A** because its a more reliable solution.

liga-oz:
closes #668

### DIFF
--- a/java-api/src/main/java/com/sap/cloud/security/token/Token.java
+++ b/java-api/src/main/java/com/sap/cloud/security/token/Token.java
@@ -8,26 +8,22 @@ package com.sap.cloud.security.token;
 import com.sap.cloud.security.config.Service;
 import com.sap.cloud.security.json.JsonObject;
 import com.sap.cloud.security.json.JsonParsingException;
-import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.security.Principal;
+import java.security.ProviderException;
 import java.time.Instant;
 import java.util.*;
 
-import static com.sap.cloud.security.token.TokenClaims.AUDIENCE;
-import static com.sap.cloud.security.token.TokenClaims.AUTHORIZATION_PARTY;
-import static com.sap.cloud.security.token.TokenClaims.ISSUER;
+import static com.sap.cloud.security.token.TokenClaims.*;
 
 /**
  * Represents a JSON Web Token (JWT).
  */
 public interface Token extends Serializable {
-
-	Logger LOGGER = LoggerFactory.getLogger(Token.class);
 
 	@SuppressWarnings("unchecked")
 	List<TokenFactory> services = new ArrayList() {
@@ -48,8 +44,10 @@ public interface Token extends Serializable {
 		if (services.isEmpty()) {
 			throw new ProviderNotFoundException("No TokenFactory implementation found in the classpath");
 		}
-		if (services.size() > 1) {
-			LOGGER.error("More than 1 service provider found, will return first one from the list");
+		if (services.size() > 2) {
+			throw new ProviderException("More than 2 TokenFactory service providers found. There should be only one");
+		}
+		if (services.size() == 2) {
 			return services.stream()
 					.filter(tokenFactory -> !tokenFactory.getClass().getName()
 							.equals("com.sap.cloud.security.servlet.HybridTokenFactory"))
--- a/java-api/src/main/java/com/sap/cloud/security/token/Token.java
+++ b/java-api/src/main/java/com/sap/cloud/security/token/Token.java
@@ -33,6 +33,8 @@ public interface Token extends Serializable {
 		}
 	};
 
+	String DEFAULT_TOKEN_FACTORY = "com.sap.cloud.security.servlet.HybridTokenFactory";
+
 	/**
 	 * Creates a token instance based on TokenFactory implementation.
 	 * 
@@ -50,7 +52,7 @@ public interface Token extends Serializable {
 		if (services.size() == 2) {
 			return services.stream()
 					.filter(tokenFactory -> !tokenFactory.getClass().getName()
-							.equals("com.sap.cloud.security.servlet.HybridTokenFactory"))
+							.equals(DEFAULT_TOKEN_FACTORY))
 					.findFirst().get().create(jwt);
 		}
 		return services.get(0).create(jwt);
--- a/java-api/src/main/java/com/sap/cloud/security/token/Token.java
+++ b/java-api/src/main/java/com/sap/cloud/security/token/Token.java
@@ -8,6 +8,7 @@ package com.sap.cloud.security.token;
 import com.sap.cloud.security.config.Service;
 import com.sap.cloud.security.json.JsonObject;
 import com.sap.cloud.security.json.JsonParsingException;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
@@ -25,6 +26,8 @@ import static com.sap.cloud.security.token.TokenClaims.ISSUER;
  * Represents a JSON Web Token (JWT).
  */
 public interface Token extends Serializable {
+
+	Logger LOGGER = LoggerFactory.getLogger(Token.class);
 
 	@SuppressWarnings("unchecked")
 	List<TokenFactory> services = new ArrayList() {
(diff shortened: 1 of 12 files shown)


java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -26,6 +27,8 @@
  */
 public interface Token extends Serializable {
 
+	Logger LOGGER = LoggerFactory.getLogger(Token.class);
```
nenaraab, line 30: Don't use LOGGER in interfaces :-)

java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -45,6 +48,13 @@ static Token create(String jwt) {
 		if (services.isEmpty()) {
 			throw new ProviderNotFoundException("No TokenFactory implementation found in the classpath");
 		}
+		if (services.size() > 1) {
+			LOGGER.error("More than 1 service provider found, will return first one from the list");
```
nenaraab, line 52: does it mean, the error is also logged, in case one has specified a custom HttpClientFactory?
then you have two SPI, the DefaultHttpClientFactory.java  and the custom one...

nenaraab (commented):
Personally I dislike this solution as it makes the implementation too complicated... maybe we should strive for using SPIs only for custom enhancements...?

java-api/src/main/java/com/sap/cloud/security/token/Token.java
liga-oz: but then I cant fulfil the prerequisite in #668 logging error

java-api/src/main/java/com/sap/cloud/security/token/Token.java
liga-oz: > Personally I dislike this solution as it makes the implementation too complicated... maybe we should strive for using SPIs only for custom enhancements...?

then the only way I see is to educate users to set in the META-INF.services file the factory classes in the right order -> first in the file is also first in the Java List

nenaraab: the problem here is that SPI does not document the behaviour in case of multiple SPI files...

java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -45,6 +44,15 @@ static Token create(String jwt) {
 		if (services.isEmpty()) {
 			throw new ProviderNotFoundException("No TokenFactory implementation found in the classpath");
 		}
+		if (services.size() > 2) {
+			throw new ProviderException("More than 2 TokenFactory service providers found. There should be only one");
```
nenaraab, line 48: "more than one customer TokenFactory" ... the default doe not count,

token-client/src/main/java/com/sap/cloud/security/client/HttpClientFactory.java
```diff
@@ -48,6 +50,16 @@ static CloseableHttpClient create(ClientIdentity clientIdentity) throws HttpClie
 		if (services.isEmpty()) {
 			throw new ProviderNotFoundException("No HttpClientFactory service could be found in the classpath");
 		}
+		if (services.size() > 2) {
+			throw new ProviderException(
+					"More than 2 HttpClientFactory service providers found. There should be only one");
```
nenaraab: same like above:
"More than one custom HttpClientFactory service providers found...."

java-api/src/main/java/com/sap/cloud/security/token/Token.java
liga-oz: good point, thx 💯

token-client/src/main/java/com/sap/cloud/security/client/HttpClientFactory.java
```diff
@@ -48,6 +50,16 @@ static CloseableHttpClient create(ClientIdentity clientIdentity) throws HttpClie
...
+		}
+		if (services.size() == 2) {
+			return services.stream()
+					.filter(httpClientFactory -> !httpClientFactory.getClass().getName()
+							.equals("com.sap.cloud.security.client.DefaultHttpClientFactory"))
```
nenaraab, line 60: would be cool, if we have a constant for that class name as part of the DefaultHttpClientFactory class. Better for refactoring.

java-api/src/main/java/com/sap/cloud/security/token/Token.java
```diff
@@ -45,6 +44,15 @@ static Token create(String jwt) {
...
+		}
+		if (services.size() == 2) {
+			return services.stream()
+					.filter(tokenFactory -> !tokenFactory.getClass().getName()
+							.equals("com.sap.cloud.security.servlet.HybridTokenFactory"))
+					.findFirst().get().create(jwt);
```
nenaraab, line 54: would be cool, if we have a constant for that class name as part of the HybridTokenFactory class. Better for refactoring.

token-client/src/test/java/com/sap/cloud/security/client/HttpClientFactoryTest.java
```diff
@@ -0,0 +1,28 @@
...
+public class HttpClientFactoryTest {
+
+	@Test
+	public void create() {
+		CloseableHttpClient cut = HttpClientFactory.create(new ClientCredentials("clientId", "secret"));
+		assertNotNull(cut);
+
+		// Assert that custom HttpClientFactory factory has a priority over default
+		// com.sap.cloud.security.client.DefaultHttpClientFactory
+		assertFalse(cut.getClass().getName().contains("InternalHttpClient"));
```
nenaraab, line 24: this assertion is always true because there is no "InternalHttpClient" at all :-) please check

nenaraab: assertTrue(cut.getClass().getName().contains("..."));

nenaraab (changes requested):
Some further comments / change requests.
Overall looks quite good!

token-client/src/test/java/com/sap/cloud/security/client/HttpClientFactoryTest.java
liga-oz: @nenaraab `InternalHttpClient` is default class resolved when `HttpClients.createDefault()` is used, that's what we have in our 'DefaultHttpClientFactory' implementation used. I could inverse that to check for mocked class instead, but with this check we are also targeting the Default implementation from our side, so I thought it's somewhat more beneficial.

java-api/src/main/java/com/sap/cloud/security/token/Token.java
liga-oz: that wouldn't be accessible in Token class thou

java-api/src/test/java/com/sap/cloud/security/servlet/HybridTokenFactory.java
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.sap.cloud.security.servlet;
+
+import com.sap.cloud.security.token.AccessToken;
+import com.sap.cloud.security.token.Token;
+import com.sap.cloud.security.token.TokenFactory;
+import org.mockito.Mockito;
+
+public class HybridTokenFactory implements TokenFactory {
```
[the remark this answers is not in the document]
liga-oz: Actually, I can't, because then I can't test the create() method, cuz based on class name I'm filtering the SPs if I rename it, tests are failing

java-api/src/main/java/com/sap/cloud/security/token/Token.java
nenaraab: ok

token-client/src/test/java/com/sap/cloud/security/client/HttpClientFactoryTest.java
nenaraab: ok. But i wasn't able to find "InternalHttpClient" in https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/token-client/src/main/java/com/sap/cloud/security/client/DefaultHttpClientFactory.java?
Maybe I'm wrong, but i would prefer a positive check to make the test bullet proof...

nenaraab (approved):

PR #708, opened 2021-11-08
XsuaaDefaultEndpoints: provide another constructor to compensate the deprecated one

### LINKED ISSUE #707: XsuaaTokenFlows: Deprecated constructors for client credential based authentication
Hello Colleagues,
could you please provide an example for the proper usage of Example Usage of XsuaaServiceConfiguration?

The background of this request is the following:
We have developed a service which in turn is calling an instance of the Market Rates Management (MRM) service. For the MRM call we first get the access token from the uaa instance bound to the MRM service. Currently we use the following coding for this purpose:

```java
XsuaaTokenFlows tokenFlows =
        new XsuaaTokenFlows(
            new DefaultOAuth2TokenService(),
            new XsuaaDefaultEndpoints(uaaUrl),
            new ClientCredentials(clientId, clientSecret));

return tokenFlows.clientCredentialsTokenFlow().execute().getAccessToken();
```

In this solution `uaaUrl`, `clientId` and `clientSecret` are coming from `uaa` environment variables (VCAP_SERVICES):
```java
@Value("${vcap.services.bpi-market-rates.credentials.uaa.url}")
String uaaUrl;
```

The solution was/is working as expected but unfortunately the used constructor version of XsuaaTokenFlows is now deprecated. For the alternative constructor we need an instance of OAuth2ServiceConfigurationProperties. I have seen that XsuaaServiceConfiguration is a specialization of OAuth2ServiceConfigurationProperties. My question now is, how do we get a proper instance of XsuaaServiceConfiguration?

Seeing that the SpringBoot annotation @ConfigurationProperties("sap.security.services.xsuaa") is used I’m wondering from where the properties are really taken from and how can I get an instance which then contains the uaa-credentials of the MRM service and not the ones of my own service?

And maybe one additional remark: at the end my own service should be called from the BTP Job Scheduler - I'm not sure if this adds additional complexity for getting the MRM credentials.

Thanks in advance for your help!

Best regards,
Sven.

P.S.: If have seen that there was already [a similar question](https://github.com/SAP/cloud-security-xsuaa-integration/issues/701) raised but this doesn't answer my question above.

nenaraab:
fixes #707

Hi @liga-oz 
Alternatively we could provide a new constructor
````
XsuaaDefaultEndpoints(String baseUri, @Nullable String certUri)
````

Furthermore, we could set the default certUri by adding the  that `https://subdomain.authentication.cert.domain`

### DIFF
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
@@ -50,9 +50,8 @@ public class XsuaaDefaultEndpoints implements OAuth2ServiceEndpointsProvider {
 	 *            - the base URI of XSUAA. Based on the base URI the tokenEndpoint,
 	 *            authorize and key set URI (JWKS) will be derived.
 	 * @param certUri
-	 *            - the cert URI
-	 *            of XSUAA. It is required in case of X.509 certificate based
-	 *            authentication.
+	 *            - the cert URI of XSUAA. It is required in case of X.509
+	 *            certificate based authentication.
 	 */
 	public XsuaaDefaultEndpoints(String baseUri, @Nullable String certUri) {
 		assertNotNull(baseUri, "XSUAA base URI must not be null.");
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
@@ -53,7 +53,7 @@ public class XsuaaDefaultEndpoints implements OAuth2ServiceEndpointsProvider {
 	 *            - the cert URI of XSUAA. It is required in case of X.509
 	 *            certificate based authentication.
 	 */
-	public XsuaaDefaultEndpoints(String baseUri, @Nullable String certUri) {
+	public XsuaaDefaultEndpoints(@Nonnull String baseUri, @Nullable String certUri) {
 		assertNotNull(baseUri, "XSUAA base URI must not be null.");
 		LOGGER.debug("Xsuaa default service endpoint: base url = {}, (cert url = {})", baseUri, certUri);
 		this.baseUri = URI.create(baseUri);
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
@@ -11,6 +11,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.net.URI;
 
 import static com.sap.cloud.security.xsuaa.Assertions.assertNotNull;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by liga-oz — CHANGES_REQUESTED** [review]
As discussed, let's keep the deprecated constructors deprecated and introduce new constructor with arguments uri and certuri(@Nullable)

File: token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java [hunk 1]
```diff
@@ -32,16 +33,32 @@
...
 		this.baseUri = baseUri;
 		this.certUri = null;
 	}
 
+	/**
+	 * Creates a new XsuaaDefaultEndpoints.
+	 *
+	 * @param baseUri
+	 *            - the base URI of XSUAA. Based on the base URI the tokenEndpoint,
+	 *            authorize and key set URI (JWKS) will be derived. - the cert URI
```
liga-oz, line 51: use @param certUri for the other argument

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
```diff
@@ -32,16 +33,32 @@
...
+	 *            of XSUAA. It is required in case of X.509 certificate based
+	 *            authentication.
+	 */
+	public XsuaaDefaultEndpoints(String baseUri, @Nullable String certUri) {
```
liga-oz, line 55: You can also add annotation `@Nonnull` in front of baseUri arg for clarity

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
```diff
@@ -32,16 +33,32 @@
...
+		assertNotNull(baseUri, "XSUAA base URI must not be null.");
+		LOGGER.debug("Xsuaa default service endpoint = {}, (cert url = {})", baseUri, certUri);
```
liga-oz, line 57: I would add here for clarity and better searchability baseUri: LOGGER.debug("Xsuaa default service endpoint: base url= {}, (cert url = {})"

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpointsTest.java
```diff
@@ -114,6 +121,6 @@ public void withQueryParametersAndEndingPathDelimiter() {
 	}
```
liga-oz, line 122: do we have a test case which covers the null check -> asserts that the exception is thrown when baseUri is null? If it's not covered would be good to have it here.

liga-oz (changes requested):
This looks so much better, well done 👍🏻

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
nenaraab: good catch !

liga-oz (approved):
💯

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpointsTest.java
nenaraab: Done

PR #710, opened 2021-11-09
fix misuse of OAuth2ServiceConfiguration.getClientIdentity() in XsuaaServiceConfiguration

liga-oz:
To avoid confusion when exception `java.lang.UnsupportedOperationException: getProperty method` is thrown for custom `XsuaaServiceConfiguration` implementations where the default `OAuth2ServiceConfiguration.getClientIdentity()` is not overridden, provide default `getClientIdentity()` method in `XsuaaClientConfiguration` interface that will notify user to implement the method and won't use the default from `OAuth2ServiceConfiguration` interface that calls unsupported `getProperty()` method.


nenaraab (approved):
💯

PR #713, opened 2021-11-10
make sure that escape characters are handled with JSON parser

nenaraab:
Currently identity service provides token issuer with special escape characters, like here:
```json
{
    "iss": "https:\/\/aoxk2addh.accounts400.ondemand.com",
    "azp": "49363701-0c40-4f12-bc3b-46d969c0ea5c",
    ...
}
```
This PR adds a test in order to make sure, that JSON Parser can handle this properly.

### DIFF
--- a/env/src/test/java/com/sap/cloud/security/json/DefaultJsonObjectTest.java
+++ b/env/src/test/java/com/sap/cloud/security/json/DefaultJsonObjectTest.java
@@ -27,7 +27,6 @@ public class DefaultJsonObjectTest {
 	private static final String STRING_VALUE = "\"" + STRING_TEXT + "\"";
 
 	private static final String STRING_LIST_VALUE = "[\"a\", \"b\", \"c\"]";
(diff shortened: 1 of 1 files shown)


strehle:
Hi , I dont see this a special behaviour so that you should take action. I assume you use a standard json library .e.g. json/jackson/json-smart etc... and they behave according to https://www.json.org/json-en.html . In this context all control characters need to be encoded , e.g. \n otherwise you have corrupted json payload. 
The / (forward slash) is special because it does not break any payload and therefore you may omit any escaping, however the / is in the list of characters to be escaped.

So I see only an issue if you have an own implemented JSON parser in place and you "forget to handle the de-escaping", e.g.
https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped .

Btw. in tools which show content of a JWT, e.g. https://jwt.io, https://jwt.ms/ you dont see this escaping .
So you can add this test but as long as you use Java and standard libraries for JSON parsing I assume you wont see any issue

liga-oz (approved):

PR #727, opened 2021-11-19
fix 'key' attribute name mismatch

liga-oz:
fix issue when `@ConfigurationProperties` doesn't load private key due to attribute name mismatch. In configuration files X509 private key is with 'key' attribute name

rename privateKey -> key

https://jtrack.wdf.sap.corp/browse/NGPBUG-168786

### DIFF
--- a/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaCredentials.java
+++ b/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaCredentials.java
@@ -6,10 +6,12 @@
 package com.sap.cloud.security.xsuaa;
 
 import com.sap.cloud.security.config.CredentialType;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/XsuaaCredentials.java
```diff
@@ -6,10 +6,12 @@
 package com.sap.cloud.security.xsuaa;
 
 import com.sap.cloud.security.config.CredentialType;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
  * Represents the XSUAA credentials of VCAP_SERVICES.
  */
+@ConfigurationProperties
```
nenaraab, line 14: don't introduce @ConfigurationProperties here

nenaraab (changes requested):
next to the removement of spring dependency in XsuaaCredentials also add CHANGELOG

nenaraab (approved):

PR #730, opened 2021-11-22
Introduce Certificate interface

liga-oz:
- Encapsulates the java.security.cert.X509Certificate

- migration and refactoring


java-security/src/main/java/com/sap/cloud/security/x509/X509CertificateExtractor.java
```diff
@@ -39,17 +37,10 @@ public static X509CertificateExtractor getInstance() {
 	 * @return the client certificate object
 	 */
 	@Nullable
-	public X509Certificate getClientCertificate(HttpServletRequest request) {
+	public String getClientCertificate(HttpServletRequest request) {
```
nenaraab, line 40: this somewhat makes the X509CertificateExtractor kind of obsolete, maybe you can put it into the AbstractAutheniticator implementation as it was used in both Authenticators

nenaraab (changes requested):

nenaraab (approved):

PR #770, opened 2021-12-21
Fix CURL user notation instead of "inline to URL"

eaglerainbow:
Most CURL do not support the `https://username:password@hostname` notation. Instead the parameter `-u` must be used. See also https://linux.die.net/man/1/curl on that.


liga-oz (approved):
thx for contribution 👍🏻

PR #787, opened 2022-01-05
Support oidc tokens from single tenant apps w/o zone_uuid claim

nenaraab:
In case of single-tenant applications ``iss`` claim in oidc token claim matches the identity provider url (``OAuth2ServiceConfiguration.getUrl())``. In that case ``x-zone_uuid`` header is not required to request the ``jwks_uri`` endpoint to retrieve token keys.


liga-oz (approved):
lgtm 👍🏻

PR #803, opened 2022-01-28
Support uaa token validation - NGPBUG-175120

nenaraab:


java-security/src/test/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticatorTest.java
```diff
@@ -105,13 +104,29 @@ public void validateIasToken_WhenRestClientIsNull() throws Exception {
...
 				response[0].getUnauthenticatedReason());
 	}
 
+	@Test
+	public void validateUaaToken() {
+		HttpServletRequest httpRequest = createRequestWithToken(uaaToken.getTokenValue());
+
+		cut = new XsuaaTokenAuthenticator()
+				.withHttpClient(mockHttpClient)
+				.withServiceConfiguration(oAuth2ServiceConfigBuilder.withClientId("dashboard_client-Id").build());
```
mwdb, line 120: should there also be a issuer?

mwdb: Git it. That is in "@Before"

mwdb (approved):

java-security/src/test/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticatorTest.java
nenaraab: oAuth2ServiceConfig/VCAP_Services is setup here:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/6bb1ef3342a040a6e62a8cc6ebe09de6fef1f6a2/java-security/src/test/java/com/sap/cloud/security/servlet/XsuaaTokenAuthenticatorTest.java#L74-L80
the token of course contains an issuer, which is 
```
"iss": "http://localhost:62700",
```
which is different to what we have in the service configuration, but in XSUAA the "Issuer-Validation" is performed on jku header (instead of iss claim) as issuer is not used.

nenaraab: see also here:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJkuValidator.java

PR #807, opened 2022-02-02
Spring Autoconfiguration Improvements

nenaraab:
Improve Autoconfig:
- [spring-security] should also auto-configure mtls based rest operations w/o credential-type=x509 property
- [spring-security] should also auto-configure token flows if no secret but certificate is given
- [spring-xsuaa] should also auto-configure mtls based rest operations w/o credential-type=x509 property


spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java
```diff
@@ -70,7 +70,7 @@ public void configures_xsuaaRestTemplate() {
 	}
 
 	@Test
-	public void configures_xsuaaMtlsRestTemplate() {
+	public void configures_xsuaaMtlsRestTemplateWithCredType() {
```
liga-oz, line 73: I think this test doesn't make any sense any longer, because even without credential-type it would work same like the test below `configures_xsuaaMtlsRestTemplate()` as the condition NoClientCertificateCondition evaluates to false. What would make sense instead would be cases where this condition evaluates to true.

liga-oz (changes requested):
I'm not sure if it makes sense in spring-xsuaa autoconfig include also the clientid check in condition as later it might lead to NP errors when [initializing tokenFlows](https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaTokenFlowAutoConfiguration.java#L65).

There are test cases that seem to be redundant:
- [serviceConfigurationDisabled_DisableDefaultPropertySourceProperty_mtls](https://github.com/SAP/cloud-security-xsuaa-integration/blob/c060adcb844a4316e09c3d0e88e6aad686cc648a/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java#L117)
- [serviceConfigurationDisabled_DisableDefaultPropertySourceProperty_nonMtls() ](https://github.com/SAP/cloud-security-xsuaa-integration/blob/c060adcb844a4316e09c3d0e88e6aad686cc648a/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java#L125)

Include test cases where `OnSecretCredentialTypeCondition` condition evaluates to true in `spring-xsuaa` module.

nenaraab:
> I'm not sure if it makes sense in spring-xsuaa autoconfig include also the clientid check in condition as later it might lead to NP errors when [initializing tokenFlows](https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaTokenFlowAutoConfiguration.java#L65).

The initialization of `XsuaaTokenFlows` seems to be quite relaxed...There is already a test which tests auto-configuration of token flows w/o given client credentials: https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaTokenFlowAutoConfigurationTest.java#L45-L52 

So, before updating that test case in `XsuaaTokenFlowAutoConfigurationTest`
````
@Test
public void autoConfigurationSkipped_if_noClientIdIsGiven() {
contextRunner.run((context) -> {
	assertThat(context).doesNotHaveBean("xsuaaTokenFlows");
});
} 
`````
we should ensure not to introduce an incompatibility... maybe the client credentials can be set at a later point in time?

> There are test cases that seem to be redundant:
> 
> * [serviceConfigurationDisabled_DisableDefaultPropertySourceProperty_mtls](https://github.com/SAP/cloud-security-xsuaa-integration/blob/c060adcb844a4316e09c3d0e88e6aad686cc648a/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java#L117)
> * [serviceConfigurationDisabled_DisableDefaultPropertySourceProperty_nonMtls() ](https://github.com/SAP/cloud-security-xsuaa-integration/blob/c060adcb844a4316e09c3d0e88e6aad686cc648a/spring-xsuaa/src/test/java/com/sap/cloud/security/xsuaa/autoconfiguration/XsuaaAutoConfigurationTest.java#L125)

Like to keep all test cases for compatibility reason, even if they are kind of redundant, when knowing the code...

> Include test cases where `OnSecretCredentialTypeCondition` condition evaluates to true in `spring-xsuaa` module.
Yes, I will check on that 👍

**Comment by nenaraab** [comment]
> > Include test cases where `OnSecretCredentialTypeCondition` condition evaluates to true in `spring-xsuaa` module.
> > Yes, I will check on that 👍

I've went through all test cases and the `OnSecretCredentialTypeCondition` and `NoClientCertificateCondition` is already covered with `configures_xsuaaRestTemplate` (existing) and with updated `configures_xsuaaMtlsRestTemplateWithCredType` and with new `configures_xsuaaMtlsRestTemplate`

**Review by liga-oz — CHANGES_REQUESTED** [review]
> I've went through all test cases and the `OnSecretCredentialTypeCondition` and `NoClientCertificateCondition` is already covered with `configures_xsuaaRestTemplate` (existing) and with updated `configures_xsuaaMtlsRestTemplateWithCredType` and with new `configures_xsuaaMtlsRestTemplate`

`configures_xsuaaRestTemplate` It only checks case with credentialType == null option, not the other 2 defined in the condition, they are not covered anywhere.
```java
return credentialType == CredentialType.BINDING_SECRET || credentialType == CredentialType.INSTANCE_SECRET
					|| credentialType == null;
```

nenaraab:
Ok. Honestly I had no focus on that test case as part of this PR as I’ve not introduced or changed this condition as part of this PR.

So my assumption is that I’ve also not reduced the code coverage.

In that case, I would suggest to add the suggested test cases as part of another PR.

liga-oz (approved):
👍🏻

PR #816, opened 2022-02-07
Fix mvn Test Run with System Stubs

nenaraab:
- [spring-xsuaa] fixes mvn test run, by using another System Stub and jupiter
- [spring-security] formats jwtDecoderBuilder


liga-oz (approved):
🥇

PR #847, opened 2022-03-22
[spring-security] Introduces spring-xsuaa compatibility module with XsuaaTokenComp

nenaraab:
Introduces a compatibility module that provides with ``XsuaaTokenComp`` class an option to decorate the token issued by xsuaa.
```xml
<dependency>
    <groupId>com.sap.cloud.security.xsuaa</groupId>
    <artifactId>spring-security-compatibility</artifactId>
    <version>2.11.15</version>
</dependency>
```

``XsuaaTokenComp`` implements `com.sap.cloud.security.xsuaa.token.Token` except for the `UserDetails` methods such as `getAuthorities()` method. Furthermore, `createInstance()` raises an `IllegalArgumentException` exception in case its not a token issued by xsuaa. 

```java
import com.sap.cloud.security.spring.token.SpringSecurityContext; // new
import com.sap.cloud.security.xsuaa.token.Token; // from spring-xsuaa

Token xsuaaToken = XsuaaTokenComp.createInstance(SpringSecurityContext.getAccessToken());
xsuaaToken.getCloneServiceInstanceId();
```

````java
import com.sap.cloud.security.token.Token;

@GetMapping("/comp/sayHello")
public Map<String, String> sayHello_compatibility(@AuthenticationPrincipal Token token) {
      com.sap.cloud.security.xsuaa.token.Token compToken;
      try {
          compToken = XsuaaTokenComp.createInstance(token);
      } catch (IllegalArgumentException e) {
          throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
      }
     compToken.getSubdomain();
     ...
}
````

**Migration Guide**<br>
https://github.com/SAP/cloud-security-xsuaa-integration/blob/token-compatibility/spring-security/Migration_SpringXsuaaProjects.md

**Open Todos**
- [x] test e2e
- [x] increase unit test coverage
- [x] add module to parent pom

### CODE REVIEW

File: spring-security/src/main/java/com/sap/cloud/security/token/XsuaaTokenComp.java [hunk 1]
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Collection;
+import java.util.Date;
+import java.util.Optional;
+
+/**
+ * Decorates a {@code Token} issued by xsuaa to provide compatibility methods for spring-xsuaa's {@code Token} interface.
+ * @deprecated please use the methods exposed by the {@code Token} interface.
+ */
+@Deprecated
+public class XsuaaTokenComp {
```
liga-oz, line 21: Not sure why you reverted renaming commit of XsuaaTokenAdapter, but I think it would be a better fitting name for this class.

nenaraab: I've undone the rename to `XsuaaTokenAdapter`, as I felt `Adapter` is strange in a class that provides Token Claim accessor methods...

```java
XsuaaTokenAdapter xsuaaTokenAdapter = XsuaaTokenAdapater.createInstance(SpringSecurityContext.getToken());
```

Do you have another idea for renaming?

liga-oz: From theoretical point of view adapter fits quite well into the game as this class does it was it says it adapts or bridges 2 incompatible classes together. Another idea XsuaaTokenWrapper, but i tend to favor the adapter more. Maybe XsuaaTokenBridge?

nenaraab: @j-denner what would be your favoured class name?

j-denner: @nenaraab I assume `XsuaaTokenComp` is for migration scenarios where there is still a lot of code using `com.sap.cloud.security.xsuaa.token.XsuaaToken`. Then the application may switch to `XsuaaTokenComp` until complete migration.

For this purpose, the following signatures are incompatible:
```Java

	@Nullable
	String getOrigin() { // This is public in com.sap.cloud.security.xsuaa.token.XsuaaToken
		return token.getClaimAsString(TokenClaims.XSUAA.ORIGIN);
	}

// ...

	public String getExpirationDate() { // Date is returned in com.sap.cloud.security.xsuaa.token.XsuaaToken
		return token.getExpiration() != null ? Date.from(token.getExpiration()).toString() : null;
	}

```

The `XsuaaToken` class extends `Jwt` and implements `com.sap.cloud.security.xsuaa.token.Token` which are both not present in `XsuaaTokenComp`.
Scenarios with `Jwt` or `com.sap.cloud.security.xsuaa.token.Token` still would break.

What about providing a separate maven artifact (xsuaa-fallback / xsuaa-compatibility) with a slightly modified `com.sap.cloud.security.xsuaa.token.XsuaaToken` where the constructor adapts from AccessToken to XsuaaToken?
(+ the required `com.sap.cloud.security.xsuaa.token.Token` dependency)

Example:
```Java
AccessToken token = SpringSecurityContext.getAccessToken();
XsuaaToken xsuaaToken = XsuaaToken.createInstance(token);
```

Then you can preserve the old API without introducing API changes and you don't get legacy code in the `spring-security` artifact.
There would be no major refactoring effort and only the construction need to be adapted (as in the current PR).
The separate module can be removed by the consumers after completing the migration.

Coming back to the original question...

I think it is not an adapter because it is an incompatible type (other inheritance, different name, different package). Secondly, you typically don't use an adapter as API. It operates as an implementation for a given interface.
`XsuaaTokenComp` is an "on method signature compatible" class.

My proposal is to go with a separate compatibility module.
If not, then I tend to something like `XsuaaTokenComp` or (less nice) `XsuaaTokenSubst` or (less nice) `XsuaaTokenAlias`.

nenaraab: @j-denner thanks a lot for your review and ideas!
- fixed `getOrigin()` -> set it to public
- fixed`getExpirationDate()` set return type to `Date`

nenaraab: My initial idea was to provide this `XsuaaTokenComp` to mitigate all common methods from `spring-xsuaa`s `com.sap.cloud.security.xsuaa.token.Token` interface, which are provided with [`SpringSecurityContext.getToken()`](https://github.com/SAP/cloud-security-xsuaa-integration/blob/token-compatibility/spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/SpringSecurityContext.java#L34).

Only in case the `AccessToken` is not enough, they could use `XsuaaTokenComp`.

+ with introducing a separate module we could implement spring-xsuaa`s Token interface, and if located in a different package, we could also name it XsuaaToken. This might simplify the initial migration step further.
- On the other hand i have to introduce up to 6 modules in order to support tomcat 10 version and I don't like to blow up the project too much...
- And having same name XsuaaToken might lead to confusions, when reading code (without having a look at the package).

I've double-checked, that I've implemented now all methods from `com.sap.cloud.security.xsuaa.token.Token` interface except for those, I've documented [here](https://github.com/SAP/cloud-security-xsuaa-integration/blob/token-compatibility/spring-security/Migration_SpringXsuaaProjects.md#token-methods).

j-denner: > * On the other hand i have to introduce up to 6 modules in order to support tomcat 10 version and I don't like to blow up the project too much...

I was **only** talking about `XsuaaToken` and it's dependencies as a compatibility layer. Therefore I do not see this dependency bloat and the dependency towards tomcat.

> * And having same name XsuaaToken might lead to confusions, when reading code (without having a look at the package).

Well maybe, it would replace the original definition for `XsuaaToken`.  As said, it's a  _compatibility_ solution (excluding construction) to reduce migration effort to an absolute minimum.

nenaraab: > I was only talking about XsuaaToken and it's dependencies as a compatibility layer. Therefore I do not see this dependency bloat and the dependency towards tomcat.

Of course, but this would then be another additional module.

> Well maybe, it would replace the original definition for XsuaaToken. As said, it's a compatibility solution (excluding construction) to reduce migration effort to an absolute minimum.

Yes, but this would delay the efforts to a later point in time.

nenaraab: @liga-oz shall we keep it as it is or would you prefer to have an extra module as @j-denner proposed?

j-denner: > > I was only talking about XsuaaToken and it's dependencies as a compatibility layer. Therefore I do not see this dependency bloat and the dependency towards tomcat.
> 
> Of course, but this would then be another additional module.

I assumed this module would be instead of `com.sap.cloud.security.xsuaa:spring-xsuaa`.

> 
> > Well maybe, it would replace the original definition for XsuaaToken. As said, it's a compatibility solution (excluding construction) to reduce migration effort to an absolute minimum.
> 
> Yes, but this would delay the efforts to a later point in time, because ideally non of these compatibility methods are ever called / used.

The `XsuaaTokenComp` class is the enabler to delay the efforts to a later point in time. The proposed approach tries to get a higher compatibility. With a separate module, it would be an explicit decision to add this layer.

liga-oz: > @liga-oz shall we keep it as it is or would you prefer to have an extra module as @j-denner proposed?

I tend to prefer the `XsuaaTokenComp` approach, as then we would need to maintain only one class instead of one module in the project

nenaraab: The above assumption that there is no rename and therefore import change isn't necessary with new module wasn't right. Applications that leverage `spring-xsuaa` mostly uses `Token` interface from `com.sap.cloud.security.xsuaa.token.Token` package. See also: 
https://github.com/SAP/cloud-security-xsuaa-integration/blob/main/samples/spring-security-xsuaa-usage/src/main/java/sample/spring/xsuaa/TestController.java#L63-L81

liga-oz (approved):

spring-security/src/main/java/com/sap/cloud/security/token/XsuaaTokenComp.java
j-denner: > The above assumption that there is no rename and therefore import change isn't necessary with new module wasn't right. Applications that leverage `spring-xsuaa` mostly uses `Token` interface from `com.sap.cloud.security.xsuaa.token.Token` package. 

@nenaraab: I don't understand this statement. I explicitly mentioned `com.sap.cloud.security.xsuaa.token.Token` to be part of the module.

My statement from above:
> What about providing a separate maven artifact (xsuaa-fallback / xsuaa-compatibility) with a slightly modified 
> com.sap.cloud.security.xsuaa.token.XsuaaToken where the constructor adapts from AccessToken to XsuaaToken?
> **(+ the required com.sap.cloud.security.xsuaa.token.Token dependency)**

Th only thing that would differ is the construction. I.e. all base classes and interfaces stay unchanged.

nenaraab: @j-denner thanks for revisiting and clarification. I got your point and will give it a try 👍

PR #849, opened 2022-03-23
Introduce Timeouts and pooling Connection Manager per client 

nenaraab:
### Motivation

Usages of HTTP Client is depicted here:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/improve-http-client/docs/images/HttpClient.drawio.svg

This improves the default http client configuration, provided with `DefaultHttpClientFactory`, so that the WARN can be removed, and stakeholders must not overwrite the default configuration.

### Details
It sets
- connect timeout = 5 sec
- connection request timeout = 5 sec
- socket timeout = 5 sec

Furthermore, it makes sure that per client id `SSLContext`, `SSLConnectionSocketFactory` and `PoolingHttpClientConnectionManager` is created only once per instance.

It introduces a `PoolingHttpClientConnectionManager` and limits
- maximum connections per route to 4 (default is 2)
- and the maximum connections to 20

With that the warning message described [here](https://github.com/SAP/cloud-security-xsuaa-integration/tree/main/token-client#new-warning-in-productive-environment-provide-well-configured-httpclientfactory-service) is no longer logged in case of certificate based setup.
In case there is no certificate given in `VCAP_SERVICES` then a default http client gets created (`HttpClients.createDefault()`) and the message is still logged with severity `WARNING` .

### Performance Tests
It uses /samples/java-token-client application and calls `/hello-token-client` which itself performs a client-credentials token flow. 

As the cache is used by default, the results are quite similar:
````
com.sap.cloud.security.test.performance.TokenClientPerformanceIT - Token client: 
Benchmark result: 1000 runs in 41462 milliseconds -> 24.12 runs/second
````

And with **disabled Token cache**, it also looks quite similar:
````
com.sap.cloud.security.test.performance.TokenClientPerformanceIT - Token client: 
Benchmark result: 1000 runs in 151050 milliseconds -> 6.62 runs/second
````

#### Referenced Issues
- #810
- #842

### CODE REVIEW

**Review by liga-oz — APPROVED** [review]

File: token-client/src/main/java/com/sap/cloud/security/client/DefaultHttpClientFactory.java [hunk 1]
```diff
@@ -24,27 +34,65 @@
...
 	public CloseableHttpClient createClient(ClientIdentity clientIdentity) throws HttpClientException {
-		LOGGER.warn("In productive environment provide well configured HttpClientFactory service");
-		if (clientIdentity != null && clientIdentity.isCertificateBased()) {
-			LOGGER.debug("Setting up HTTPS client with: certificate: {}\n", clientIdentity.getCertificate());
-			SSLContext sslContext;
+		if (clientIdentity != null) {
+			if (httpClientsCreated.contains(clientIdentity.getId())) {
+				LOGGER.warn("Application has already created HttpClient for {}, please check.", clientIdentity.getId());
+			}
+			httpClientsCreated.add(clientIdentity.getId());
```
j-denner, line 59: This code is not thread safe. A simple solution would be to make `httpClientsCreated` a HashSet and synchronize it with `Collections.synchronizedSet(...)`.

However, the overall flow does not capture the case when `clientIdentity == null` but still HttpClients get created.

nenaraab: done 👍

PR #855, opened 2022-03-28
Leverage BTP environment accessor lib

nenaraab:
Uses https://github.com/SAP/btp-environment-variable-access (version 0.3.0), which supports access to service credentials provisioned by [SAP BTP Service Operator](https://github.com/SAP/sap-btp-service-operator). SAP BTP Service Operator is the successor of Service Bindings managed by Service Catalog, which is deprecated.

Service-manager is no longer required to distinguish by plan, if multiple xsuaa instances are bound.

Extract of deployment.yml:

```yml
apiVersion: services.cloud.sap.com/<version>
kind: ServiceInstance
metadata:
  name: auth
spec:
  serviceOfferingName: identity
  servicePlanName: application
  parameters:
    authorization:
      product_label: "IAS-AMS Spring example app (Kyma)"

---
apiVersion: services.cloud.sap.com/<version>
kind: ServiceBinding
metadata:
  name: identity
spec:
  serviceInstanceName: auth
  parameters:
    credential-type: "X509_GENERATED"
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: ams-sample-app
  name: ams-sample-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: ams-sample-app
  strategy: {}
  template:
    metadata:
      labels:
        app: ams-sample-app
    spec:
      imagePullSecrets:
        - name: sap-image-registry
      containers:
        - image: <YOUR IMAGE REPOSITORY>
          imagePullPolicy: Always
          name: backend
          ports:
            - name: http
              containerPort: 8080
          volumeMounts:
            - name: auth-secret
              mountPath: "/bindings/identity"
              readOnly: true
          env:
            - name: SERVICE_BINDING_ROOT
              value: "/bindings"
```

### Further References
- https://github.com/SAP/sap-btp-service-operator


java-api/src/main/java/com/sap/cloud/security/config/Service.java
```diff
@@ -5,9 +5,12 @@
  */
 package com.sap.cloud.security.config;
 
+import com.sap.cloud.security.token.GrantType;
```
Johannes-Schneider, line 8: Unused import.

```suggestion
```

env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -48,15 +55,58 @@ public Type getType() {
...
+	}
+
+	private OAuth2ServiceConfiguration mapToOAuth2ServiceConfiguration(ServiceBinding b) {
+		if (!b.getServiceName().isPresent()) {
+			LOGGER.error("Ignores Service Binding with name {} as service name is not provided.", b.getName());
+			return null; // TODO test
+		}
+		final Service service = from(b.getServiceName().get());
+		OAuth2ServiceConfigurationBuilder configBuilder =  OAuth2ServiceConfigurationBuilder.forService(service)
+				.withProperties(TypedMapView.ofCredentials(b).getEntries(String.class))
```
Johannes-Schneider, line 78: > `TypedMapView.ofCredentials(b).getEntries(String.class)`

I fear this might not be doing what you would expect it to do...
This method will **filter** the underlying `Map<String. Object>` (the "raw" credentials) for entries that match the queried type (`String` in this case).
All other entries (such as numbers, booleans, Lists and complex Objects) will **not** be returned.

----

I'm generally not so happy about how the `TypedMapView` needs to be used here.
I think we should probably add some compatibility methods such as `.toMap` and `.toList` to the API.
Otherwise things are just too cumbersome.

What do you think, @nenaraab?

env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -48,15 +55,58 @@ public Type getType() {
...
 
+	private void loadAll() {
+		Map<String, OAuth2ServiceConfiguration> xsuaaPlans = DefaultServiceBindingAccessor.getInstance().getServiceBindings().stream()
+				.filter(b -> XSUAA.getCFName().equalsIgnoreCase(b.getServiceName().orElse(null)))
+				.map(b -> mapToOAuth2ServiceConfiguration(b)).collect(Collectors.toMap(c -> c.getProperty("plan"),
+						Function.identity()));
+		Map<String, OAuth2ServiceConfiguration> identityPlans = DefaultServiceBindingAccessor.getInstance().getServiceBindings().stream()
+				.filter(b -> IAS.getCFName().equalsIgnoreCase(b.getServiceName().orElse(null)))
+				.map(b -> mapToOAuth2ServiceConfiguration(b)).collect(Collectors.toMap(c -> c.getProperty("plan"),
+						Function.identity()));
```
Johannes-Schneider, line 66: You could consider not loading the bindings twice:

```suggestion
		List<ServiceBinding> serviceBindings = DefaultServiceBindingAccessor.getInstance().getServiceBindings();
		Map<String, OAuth2ServiceConfiguration> xsuaaPlans = serviceBindings.stream()
				.filter(b -> XSUAA.getCFName().equalsIgnoreCase(b.getServiceName().orElse(null)))
				.map(this::mapToOAuth2ServiceConfiguration).collect(Collectors.toMap(c -> c.getProperty("plan"),
						Function.identity()));
		Map<String, OAuth2ServiceConfiguration> identityPlans = serviceBindings.stream()
				.filter(b -> IAS.getCFName().equalsIgnoreCase(b.getServiceName().orElse(null)))
				.map(this::mapToOAuth2ServiceConfiguration).collect(Collectors.toMap(c -> c.getProperty("plan"),
						Function.identity()));
```

java-api/src/main/java/com/sap/cloud/security/config/Service.java
nenaraab: thanks 👍

env/pom.xml
```diff
@@ -23,17 +23,17 @@
...
             <version>${project.version}</version>
         </dependency>
-        <!-- Required only in case of multiple xsuaa bindings -->
         <dependency>
-            <groupId>com.sap.cloud.security.xsuaa</groupId>
-            <artifactId>token-client</artifactId>
-            <version>${project.version}</version>
-            <scope>provided</scope>
+            <groupId>com.sap.cloud.environment.servicebinding</groupId>
+            <artifactId>java-sap-service-operator</artifactId>
```
Johannes-Schneider, line 28: Just to get that straight: This module (the `com.sap.cloud.security:env`) is supposed to **only** support K8s based applications, right?

env/src/test/java/com/sap/cloud/security/config/EnvironmentsK8sTest.java
```diff
@@ -5,23 +5,40 @@
...
 
 	private static final String K8S_HOST_VALUE = "0.0.0.0";
 
-	@Test
-	void getCurrent_returnsOnlySingleK8sInstance(EnvironmentVariables environmentVariables) {
+	@BeforeAll
+	static void beforeAll(EnvironmentVariables environmentVariables) {
 		environmentVariables.set(KUBERNETES_SERVICE_HOST, K8S_HOST_VALUE);
+		DefaultServiceBindingAccessor.setInstance(new SapServiceOperatorLayeredServiceBindingAccessor(
+				Paths.get("src/test/resources/k8s"), DEFAULT_PARSING_STRATEGIES));
```
Johannes-Schneider, line 32: Hardcoding the resource path like this is prone to error IMO.
AFAIK, the best practice is to get the `ClassLoader` of this test class and then get a resource from there.

Look at [this implementation](https://github.com/SAP/btp-environment-variable-access/blob/main/sap-service-operator/src/test/java/com/sap/cloud/environment/servicebinding/TestResource.java#L16-L57) (and [the usage](https://github.com/SAP/btp-environment-variable-access/blob/main/sap-service-operator/src/test/java/com/sap/cloud/environment/servicebinding/SapServiceOperatorLayeredServiceBindingAccessorTest.java#L29)) for an example.

pom.xml
```diff
@@ -66,6 +66,7 @@
 		<log4j2.version>2.18.0</log4j2.version>
 		<slf4j.api.version>1.7.36</slf4j.api.version> <!--see also here http://www.slf4j.org/faq.html#changesInVersion18 -->
 		<org.json.version>20220320</org.json.version>
+		<sap.cloud.env.servicebinding.version>0.2.0</sap.cloud.env.servicebinding.version>
```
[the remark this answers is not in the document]
Johannes-Schneider: Seems like you were too fast for me, @nenaraab 😅

env/pom.xml
nenaraab: yes, as of now

env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -48,15 +55,63 @@ public Type getType() {
...
+		serviceConfigurations.put(XSUAA, xsuaaPlans);
+		serviceConfigurations.put(IAS, identityPlans);
+	}
+
+	private OAuth2ServiceConfiguration mapToOAuth2ServiceConfiguration(ServiceBinding b) {
+		if (!b.getServiceName().isPresent()) {
+			LOGGER.error("Ignores Service Binding with name {} as service name is not provided.", b.getName());
+			return null; // TODO test
```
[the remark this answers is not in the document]
nenaraab: oh wow thanks, will check that ;-)

env/src/test/java/com/sap/cloud/security/config/EnvironmentsK8sTest.java
nenaraab: Done with `Paths.get(K8sEnvironmentTest.class.getResource("/k8s").getPath())`

env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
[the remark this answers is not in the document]
nenaraab: this code will never be reached, added test case that demos that, but still if its called, the behaviour of the method is correct

PR #858, opened 2022-03-29
replace jackson-databind version to solve sec vulnerability

nenaraab:


liga-oz (approved):
I added the dependency definition centrally as the vulnerable version got pulled in as transitive dependency in other modules too.

PR #876, opened 2022-04-13
Dependency on spotbugs-annotations should be optional #851

### LINKED ISSUE #851: Dependency on spotbugs-annotations should be optional
**Is your feature request related to a problem? Please describe.**

The dependency on spotbugs-annotations in 
[cloud-security-xsuaa-integration](https://github.com/SAP/cloud-security-xsuaa-integration/tree/ade5764ec22ba60ecee726988bd3fb0ec1995dce)/[java-api](https://github.com/SAP/cloud-security-xsuaa-integration/tree/ade5764ec22ba60ecee726988bd3fb0ec1995dce/java-api)/pom.xml
is transitive. This means projects relying on java-api will be delivering spotbugs-annotations and findbugs:jsr305 to their clients.

I assume this is not required as these annotations are not processed at runtime.

Also the documentation at https://spotbugs.readthedocs.io/en/stable/migration.html#guide-for-migration-from-findbugs-3-0-to-spotbugs-3-1 recommends to make this dependency optional, so by default it will not propagate into downstream projects.

**Describe the solution you'd like**

The dependency on spotbugs-annotations should be optional.

**Describe alternatives you've considered**

Alternatively, the dependency could be excluded by downstream projects but this requires an effort for every project. A sane default with the "optional" dependency would be preferred.

nenaraab:
closes #851 
follows recommendation from [here](https://spotbugs.readthedocs.io/en/stable/migration.html#guide-for-migration-from-findbugs-3-0-to-spotbugs-3-1)

[all]
- Dependency on spotbugs-annotations should be optional <br>

[spring-xsuaa]
- Dependency on javax.annotation-api should be optional

### DIFF
--- a/env/pom.xml
+++ b/env/pom.xml
@@ -36,6 +36,11 @@
             <artifactId>httpclient</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+            <optional>true</optional>
+        </dependency>
 
         <!-- test utilities -->
         <dependency>
--- a/java-api/pom.xml
+++ b/java-api/pom.xml
@@ -26,6 +26,7 @@
 		<dependency>
 			<groupId>com.github.spotbugs</groupId>
 			<artifactId>spotbugs-annotations</artifactId>
+			<optional>true</optional>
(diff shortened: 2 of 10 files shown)


nenaraab:
sample builds are failing, as version 2.11.16 can't be found on maven central

PR #1063, opened 2023-01-13
Make zoneid check in JwtSignatureValidator configurable

liga-oz:


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureValidator.java
```diff
@@ -58,6 +59,19 @@ class JwtSignatureValidator implements Validator<Token> {
...
+	/**
+	 * This method disables the zone check. In case Jwt issuer `iss` claim doesn't
+	 * match with the `url` attribute from {@link OAuth2ServiceConfiguration)},
+	 * zone-id claim needs to be present in token to ensure that the zone belongs to
+	 * this issuer.
+	 * <p>
+	 * Use with caution as it relaxes the validation rules! It is not recommended to
+	 * disable this check for standard Identity service setup.
+	 */
+	void disableZoneCheck() {
```
mwdb, line 71: We still need to deprecate zones. Should we call it a bit more neutral, e.g. disableApplicationTenantIdCheck?

PR #1075, opened 2023-02-01
Switch to service binding accessor library

finkmanAtSap:
... instead of using CFEnvParser


liga-oz (changes requested):
As discussed `OAuth2ServiceConfigurationBuilder fromServiceBinding(ServiceBinding b)` needs to be removed as our internally used dependencies shouldn't be exposed to end users, the option you mentioned by adding deprecated utility class, fits the best for the scenario.

env/src/test/java/com/sap/cloud/security/config/cf/CFEnvironmentTest.java
```diff
@@ -165,6 +148,19 @@ public void getConfigurationByPlan() {
 
 	@Test
 	public void getXsuaaServiceConfiguration_usesSystemProperties() {
+		String previousSystemProperty = System.getProperty(VCAP_SERVICES, vcapMultipleXsuaa);
+		System.setProperty(VCAP_SERVICES, vcapMultipleXsuaa);
+		cut = CFEnvironment.getInstance();
+		System.setProperty(VCAP_SERVICES, previousSystemProperty);
```
liga-oz, line 154: what exactly you're testing here? in L151 if there's no VCAP services you set `previousSystemProperty` to `vcapMultipleXsuaa` then in line L152 you set VCAP_SERVICES to `vcapMultipleXsuaa` get the CfEnv instance then reset the properties again to `previousSystemProperty` which can have same value as `vcapMultipleXsuaa` and then in L159 you're asserting that the config had multiple xsuaa configs, but in both `System.setProperty(...)` it can contain value of `vcapMultipleXsuaa` 🤔 looks a bit ambiguous to me

liga-oz (changes requested):
all in all looks quite good, just a few comments.

env/src/test/java/com/sap/cloud/security/config/cf/CFEnvironmentTest.java
finkmanAtSap: Good catch! In line 151 I intended to call System.getProperty(VCAP_SERVICES) without vcapMultipleXsuaa as second argument. Seems like a copy/paste mistake.
The intention is to test if the default CFEnvironment looks into the system properties for configurations. I intend to remember and restore the value of VCAP_SERVICES after testing with my supplied value. As we are planning to remove our dependencies on the usual libs that provide this kind of remember/restore functionality, I am doing this manually.

liga-oz (approved):
lgtm

PR #1081, opened 2023-02-08
Support xs.system.attributes as string value

finkmanAtSap:
I am not really happy with the code style of the provided solution. Maybe you have an idea for a more concise solution @liga-oz?


java-security/src/main/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapter.java
```diff
@@ -405,11 +405,26 @@ void setOAuth2TokenService(OAuth2TokenService oAuth2TokenService) {
 	}
 
 	private String[] getMultiValueAttributeFromExtObject(String claimName, String attributeName) {
-		List<String> claims = accessToken.getAttributeFromClaimAsStringList(claimName, attributeName);
-		if (claims.isEmpty() && !(claims instanceof ArrayList)) {
+		List<String> values = null;
+
+		try {
+			values = accessToken.getAttributeFromClaimAsStringList(claimName, attributeName);
+		} catch (JsonParsingException e) {}
```
liga-oz, line 412: I would move the lines 416-419 into this catch block

java-security/src/test/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapterTest.java
```diff
@@ -521,6 +521,46 @@ public void isForeignModeTrue_whenIdentityZoneIsMissing() throws XSUserInfoExcep
 		assertThat(cut.isInForeignMode()).isTrue();
 	}
 
+	@Test
```
liga-oz, line 524: Here's a lot of repeated code, I would suggest to create these test cases as parameterized test and use Mocking for XsuaaToken instead of creating multiple Jwt token files in resources.

liga-oz (changes requested):
See comments added

java-security/src/test/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapterTest.java
```diff
@@ -537,28 +542,39 @@ public void canReadSystemAttributesAsPlainString() throws IOException {
...
-	public void readingSystemAttributesAsIntegerRaisesException() throws IOException {
-		token = new XsuaaToken(IOUtils.resourceToString("/tokenWithSystemPropertyAttributeAsInteger.txt", UTF_8));
-		cut = new XSUserInfoAdapter(token);
+	private static Stream<Arguments> readingInvalidSystemAttributesRaisesExpectedException() {
+		String attributeName = "foo";
 
-		assertThatThrownBy(() -> cut.getSystemAttribute("foo")).isInstanceOf(XSUserInfoException.class);
-	}
+		Token t1 = Mockito.mock(AccessToken.class);
+		when(t1.getAttributeFromClaimAsString(XS_SYSTEM_ATTRIBUTES, attributeName)).thenReturn(null);
```
liga-oz, line 557: this still is a repetitive code, it needs to be optimised plus you can also add the 2 positive test cases (canReadSystemAttributesAsStringArray, canReadSystemAttributesAsPlainString) in the same test. 
Consider initialising the token mock in the test class itself, pass expected return values of `getAttributeFromClaimAsString` and `getAttributeFromClaimAsStringList` instead of token as parameters.

liga-oz (changes requested):
Test class needs some improvements.

java-security/src/test/java/com/sap/cloud/security/adapter/xs/XSUserInfoAdapterTest.java
finkmanAtSap: I just submitted a new version that should suit your suggestions. However, the test method contains a lot of lines to differentiate and cast between the positive and negative test cases. I think two separate positive and negative test methods would be better readable. If you agree, I can split it up while keeping the improvements in regards to less repetition.

finkmanAtSap: Thanks for the suggestion! Implemented in latest commit.

liga-oz (changes requested):

PR #1125, opened 2023-03-27
Bugfix sample tests for 3.0.0

finkmanAtSap:


java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTest.java
```diff
@@ -189,46 +188,29 @@ public String getApplicationServerUri() {
...
+		context.setResourceBase("src/main/java/webapp");
 
-		return servletHandler;
-	}
+		// add Servlets
+		applicationServletsByPath
+				.forEach((path, servletHolder) -> context.addServlet(servletHolder, path));
 
-	WebAppContext createWebAppContext() {
-		WebAppContext context = new WebAppContext();
```
liga-oz, line 223: `WebAppContext` is an extension to `ServletContextHandler` and supports servlets configured in web.xml or by annotations without a need to explicitly set them in SecurityTestExtension like here:
```java 
SecurityTestExtension.forService(XSUAA)
			.useApplicationServer();
			.addApplicationServlet(HelloJavaServlet.class, HelloJavaServlet.ENDPOINT)
```
The way how you overwrote that is a simpler way and apps that use web.xml for example would need to define each servlet like mentioned above, meaning we take a way some flexibility, therefore, I would suggest we stick with the `WebAppContext`

finkmanAtSap: I was able to fix the 501 Jetty problem using WebAppContext now.

liga-oz (changes requested):
This setup doesn't seem to work with web.xml. We should adapt one of the java-security-usage samples' integration test to use servlet definitions from web.xml instead of defining the servlets in the test extension itself, smth like: 
```java
SecurityTestExtension extension = SecurityTestExtension.forService(XSUAA)
			.useApplicationServer();
//			.addApplicationServlet(HelloJavaServlet.class, HelloJavaServlet.ENDPOINT)
```
or perhaps add additional tests in the java-security-test module that covers web.xml usage.

java-security-test/src/main/java/com/sap/cloud/security/test/SecurityTest.java
```diff
@@ -188,31 +189,43 @@ public String getApplicationServerUri() {
 	}
 
 	void startApplicationServer() throws Exception {
-		// create SecurityHandler
-		JettyTokenAuthenticator authenticator = new JettyTokenAuthenticator(
-				applicationServerOptions.getTokenAuthenticator());
-		ConstraintSecurityHandler security = new ConstraintSecurityHandler();
-		security.setAuthenticator(authenticator);
+		WebAppContext context = createWebAppContext();
+		ServletHandler servletHandler = createServletHandler(context);
```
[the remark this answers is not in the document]
finkmanAtSap: It looks like an additional ServletHandler is indeed not necessary to add servlets to a WebAppContext. I removed it.

liga-oz (approved):
Nicely done 💯

Issue #1133, opened 2023-04-01
Dependency org.yaml:snakeyaml, leading to CVE problem


CVEDetect:
Hi, In **/samples/spring-security-hybrid-usage**，there is a dependency **org.yaml:snakeyaml:1.30** that calls the risk method.

[CVE-2022-25857](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-25857)

The scope of this CVE affected version is **[0,1.31)**

After further analysis, in this project, the main Api called is **org.yaml.snakeyaml.composer.Composer: composeNode(org.yaml.snakeyaml.nodes.Node)Lorg.yaml.snakeyaml.nodes.Node;**

Risk method repair link : [GitHub](https://github.com/snakeyaml/snakeyaml/commit/fc300780da21f4bb92c148bc90257201220cf174)

**CVE Bug Invocation Path--**

**Path Length : 6**

```
CVE Bug Invocation Path : 
CVE Bug Invocation Path : 
sample.spring.security.SecurityConfiguration$MyCustomHybridTokenAuthenticationConverter: convert(org.springframework.security.oauth2.jwt.Jwt)Lorg.springframework.security.authentication.AbstractAuthenticationToken; /.m2/repository/jakarta/annotation/jakarta.annotation-api/1.3.5/jakarta.annotation-api-1.3.5.jar
sample.spring.security.SecurityConfiguration$MyCustomHybridTokenAuthenticationConverter: deriveAuthoritiesFromGroup(org.springframework.security.oauth2.jwt.Jwt)Ljava.util.Collection; /.m2/repository/jakarta/annotation/jakarta.annotation-api/1.3.5/jakarta.annotation-api-1.3.5.jar
org.yaml.snakeyaml.Yaml$2: next()Ljava.lang.Object; /.m2/repository/com/fasterxml/jackson/core/jackson-core/2.13.4/jackson-core-2.13.4.jar
org.yaml.snakeyaml.Yaml$2: next()Lorg.yaml.snakeyaml.nodes.Node; /.m2/repository/com/fasterxml/jackson/core/jackson-core/2.13.4/jackson-core-2.13.4.jar
org.yaml.snakeyaml.composer.Composer: getNode()Lorg.yaml.snakeyaml.nodes.Node; /.m2/repository/com/fasterxml/jackson/core/jackson-core/2.13.4/jackson-core-2.13.4.jar
[...]

PR #1134, opened 2023-04-01
Fix CVE dependency issue

CVEDetect:
Fix issue #1133   by update dependency org.yaml:snakeyaml:1.33 @liga-oz


liga-oz:
Thanks, for you PR, but according to https://github.com/spring-projects/spring-boot/issues/33457 spring does not use the affected constructor, it uses SafeConstructor and only for application.yaml file parsing, therefore it's a false positive and with that I will close this PR.

PR #1135, opened 2023-04-03
Add JwtBearerTokenFlow to 2.x and Deprecate UserTokenFlow

MatKuhr:


token-client/src/main/java/com/sap/cloud/security/xsuaa/tokenflows/JwtBearerTokenFlow.java
```diff
@@ -83,14 +116,16 @@ public JwtBearerTokenFlow subdomain(String subdomain) {
...
      *
-     * @param optionalParameters
-     *            - the optional parameters.
+     * @param additionalAuthorizationAttributes
+     *            - the additional attributes.
      * @return this builder.
      */
-    public JwtBearerTokenFlow optionalParameters(Map<String, String> optionalParameters) {
-        this.optionalParameters = optionalParameters;
+    public JwtBearerTokenFlow attributes(Map<String, String> additionalAuthorizationAttributes) {
```
MatKuhr, line 127: @liga-oz Do you plan to also apply the same change on `main-dev`? Because otherwise we would have a breaking change here between v2 and v3, right?

token-client/src/main/java/com/sap/cloud/security/xsuaa/tokenflows/JwtBearerTokenFlow.java
```diff
@@ -41,17 +47,44 @@ public JwtBearerTokenFlow(@Nonnull OAuth2TokenService tokenService,
...
+    }
+
+    /**
+     * Sets the zoneid of the tenant<br>
+     *
+     * @param zoneId
+     *            - the zoneId.
+     * @return this builder.
+     */
+    public JwtBearerTokenFlow zoneId(String zoneId) {
```
MatKuhr, line 83: If I give a zone ID here that is `!=` the ID contained within the `Token token` passed above, what would happen? Should we throw an exception in this case?

liga-oz: if the zone id doesn't exists xsuaa would respond with 404 This `zoneId(String zoneId)` setter was meant to be used together with `token(String bearerToken)` as with this setter there's no option to set the `X-zid` header unless you use `token(Token token)` setter. I wouldn't restrict setting the zone id to the same value as in the token as I cannot enforce the same logic in case `token(String bearerToken)`  is used, moreover zone id header is still optional.

PR #1136, opened 2023-04-04
Support httpclient 4 and 5

liga-oz:


spring-security/src/main/java/com/sap/cloud/security/spring/autoconfig/XsuaaTokenFlowAutoConfiguration.java
```diff
@@ -60,59 +58,27 @@ public XsuaaTokenFlowAutoConfiguration(XsuaaServiceConfigurations xsuaaConfigs,
...
-	/**
-	 * Creates a certificate based {@link RestOperations} instance if the
-	 * application has not defined any.
+	 * Creates a {@link CloseableHttpClient} instance configured with the
+	 * ClientIdentity provided. Conditional on missing CloseableHttpClient Bean.
 	 *
-	 * @return the {@link RestOperations} instance.
+	 * @return the {@link CloseableHttpClient} instance.
 	 */
 	@Bean
```
finkmanAtSap, line 76: Should we maybe add a bean qualifier here, e.g. @Bean(name = "xsuaaHttpClient") and @ConditionalOnMissingBean(name = "xsuaaHttpClient"). Then, users of the library who need to define further beans of type CloseableHttpClient in their application, i.e. beans not intended for communication to the XSUAA service, do not run into problems?
We use type-based injection to inject an CloseableHttpClient into the method that creates the XsuaaTokenFlows bean., so I think the **user** would be forced to **qualify** any CloseableHttpClient beans he creates to prevent them from being injected into our method. If he provides further client beans without a qualifier, they would be used by us, even if intended for a different communication channel. I think it would be good manners to leave the non-qualified bean to the main application?

token-client/src/main/java/com/sap/cloud/security/client/DefaultSpringHttpClientFactory.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Override
+	public RestTemplate createRestTemplateClient(ClientIdentity clientIdentity) {
+		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
+		requestFactory.setHttpClient(createClient(clientIdentity));
+		return new RestTemplate(requestFactory);
+	}
+
+	private CloseableHttpClient createClient(ClientIdentity clientIdentity)
+			throws HttpClientException {
```
finkmanAtSap, line 48: We used to print a conditional warning when creating a client about not using this default client for production. If you think, it would still make sense, we could add it here to the beginning of the method again?

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenService.java
```diff
@@ -5,33 +5,33 @@
...
-import org.apache.hc.client5.http.entity.UrlEncodedFormEntity;
-import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
-import org.apache.hc.core5.http.HttpStatus;
-import org.apache.hc.core5.http.io.entity.EntityUtils;
-import org.apache.hc.core5.http.message.BasicNameValuePair;
-import org.json.JSONObject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static com.sap.cloud.security.xsuaa.client.OAuth2TokenServiceConstants.*;
+import static org.apache.hc.core5.http.HttpHeaders.USER_AGENT;
```
finkmanAtSap, line 9: Should be imported from client 4, not 5

token-client/src/main/java/com/sap/cloud/security/client/SpringHttpClientFactory.java
finkmanAtSap, line 1: Is it necessary to have this in token-client instead of spring-xsuaa? From what I can see, the interface is only used in spring-xsuaa and token-client uses apache http client 4 for token flows. I would suggest moving this interface (along with DefaultSpringHttpClientFactory and the META-INF) to the only module that makes use of it unless there are plans to use it again in other modules once we switch to apache http client 5 in the future.

finkmanAtSap (commented):
Please consider the comments I made before merging.

token-client/src/main/java/com/sap/cloud/security/client/DefaultSpringHttpClientFactory.java
liga-oz: I didn't add them here, because we're going to get rid of this message also in the DefaultHttpClientFactory(in separate PR) as per request from cloud sdk and let end users use the Default impl and document it, so that if they see performance degradation they have to provide their own impl. This will solve the issue when 3 implementation are provided: 1st from us, 2nd from cloud sdk and then 3rd from end user and then end user runs into this error https://github.com/SAP/cloud-security-services-integration-library/blob/main-2.x/token-client/src/main/java/com/sap/cloud/security/client/HttpClientFactory.java#L65

token-client/src/main/java/com/sap/cloud/security/client/SpringHttpClientFactory.java
liga-oz: good point 👍🏻

spring-security/src/test/java/com/sap/cloud/security/spring/autoconfig/XsuaaTokenFlowAutoConfigurationTest.java
```diff
@@ -119,43 +118,24 @@ void autoConfigurationUsesTokenFlowsForMultipleXsuaaServicesAsPrimary() {
...
-				.withPropertyValues(mt_properties.toArray(new String[0]))
-				.withConfiguration(AutoConfigurations.of(HybridIdentityServicesAutoConfiguration.class,
-						XsuaaTokenFlowAutoConfiguration.class));
-
-		mt_runner.run(context -> assertTrue(context.containsBean("xsuaaTokenFlows")));
-		mt_runner.run(context -> assertTrue(context.containsBean("mtlsRestOperations")));
-	}
-
 	@Test
 	void userConfigurationCanOverrideDefaultBeans() {
```
finkmanAtSap, line 122: I think this test only asserts that the user can override both the HttpClientFactory AND the XsuaaTokenFlows beans in tandem. The more common case which is not tested here, is probably users overriding only the http client.
If the intended way is to override both beans, we should consider adding comments or documentation for the users.
If users should be able to override only the http client bean, we should not use "customHttpClient" as bean qualifier in the test, as this will not be used (because XsuaaTokenFlows bean now has its http client injected with qualifier "tokenFlowHttpClient").

token-client/src/main/java/com/sap/cloud/security/client/DefaultHttpClientFactory.java
```diff
@@ -63,41 +58,42 @@ public CloseableHttpClient createClient(ClientIdentity clientIdentity) throws Ht
...
+			SslConnection connectionPool = sslConnectionPool.computeIfAbsent(clientId,
+					s -> new SslConnection(clientIdentity));
+			return HttpClients.custom()
+					.setDefaultRequestConfig(requestConfig)
+					.setConnectionManager(connectionPool.poolingConnectionManager)
+					.setSSLContext(connectionPool.context)
+					.setSSLSocketFactory(connectionPool.sslSocketFactory)
+					.build();
 		}
+		LOGGER.warn(
```
finkmanAtSap, line 71: Not sure if we want to remove the warning in this place as well. If we want to keep it, please note that it is currently not printed for the TLS case above. Not sure if that is intended.

finkmanAtSap (commented):
Just a few more remarks on things I am not fully confident about. Feel free to handle them according to your own judgement.

spring-security/src/test/java/com/sap/cloud/security/spring/autoconfig/XsuaaTokenFlowAutoConfigurationTest.java
liga-oz: good point, but i can't even test it as it has to be the same name for the bean, but yeah readme need to be updated 👍🏻

liga-oz: it's actually already described in readme with a link to spring boot documentation, so seems to be all good here

PR #1137, opened 2023-04-04
Simplify TokenBrokerResolver in Basic Auth Sample

finkmanAtSap:
First step to adapt sample for 3.0.0.

Reduces TokenBrokerResolver in spring-security-basic-auth sample to BasicAuth case, use XsuaaTokenFlows and remove 
related dead code.


samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/security/TokenBrokerResolver.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+import java.util.Optional;
+
+/**
+ * Resolves with an XSUAA user token from the configured service instance using the Basic HTTP credentials
```
liga-oz, line 26: how about:
Resolves an user token from the configured XSUAA service instance via password grant flow using the Basic HTTP credentials from the request header.

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/security/TokenBrokerResolver.java
```diff
@@ -0,0 +1,126 @@
...
+ * in the request header to execute a {@link com.sap.cloud.security.token.GrantType#PASSWORD} grant type flow.
+ * <p>
+ * The user crendentials are expected to be contained in the {@link HttpHeaders#AUTHORIZATION} header field as BASE64-encoded username/password combination.
+ * Optionally, a zone id may be specified via the 'X-Identity-Zone-Subdomain' header field.
+ * <p>
+ * For performance improvements, a pre-configured {@link Cache} may be supplied to limit the time between two token requests for the same principal.
+ */
+public class TokenBrokerResolver implements BearerTokenResolver {
+	private static final Logger logger = LoggerFactory.getLogger(TokenBrokerResolver.class);
```
liga-oz, line 35: private static final usually is uppercase, `logger` should be consistent with other constants you defined e.g.  BASIC_AUTHENTICATION_SCHEME

samples/spring-security-basic-auth/src/main/java/sample/spring/xsuaa/security/TokenBrokerResolver.java
```diff
@@ -0,0 +1,126 @@
...
+			cacheKey = createSecureHash(credentials.userName(), credentials.password);
+		}
+
+		String cachedToken = tokenCache.get(cacheKey, String.class);
+		if (cachedToken != null) {
+			logger.debug("returning access token for {} from cache.", cacheKey);
+			return cachedToken;
+		}
+
+		PasswordTokenFlow tokenFlow = tokenFlows
```
liga-oz, line 82: i think you don't need a separate variable for tokenFlows just merge it with the token var.
```java		
String token = tokenFlows
		.passwordTokenFlow()
		.username(credentials.userName())
		.password(credentials.password)
		.subdomain(zoneId)
		.execute()
		.getAccessToken();
```

samples/spring-security-basic-auth/src/test/java/sample/spring/xsuaa/SecurityConfigurationTest.java
liga-oz, line 1: nicely commented tests 💯

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/XsuaaDefaultEndpoints.java
```diff
@@ -36,6 +36,16 @@ public class XsuaaDefaultEndpoints implements OAuth2ServiceEndpointsProvider {
 	 *            - the cert URI of XSUAA. It is required in case of X.509
 	 *            certificate based authentication.
 	 */
+	public XsuaaDefaultEndpoints(@Nonnull URI baseUri, @Nullable URI certUri) {
```
liga-oz, line 39: this is out of scope of this PR and I don't see you using the added constructor in the  basic auth sample either, please remove it.

liga-oz (changes requested):
The SecurityConfigurationTest was the highlight of the PR, well done 🥇 
The package structure is a bit overkill for such a small sample, it doesn't improve the visibility of project's files when you have to open up 3 folders and find there at most 2 files. For example under config you have a SecurityConfiguration class, but then you also have a security folder I find it confusing.  If you really want to keep these packages, then maybe rename security to tokenbroker and move the TokenBrokerResolver and TokenBrokerConfiguration into it, but I would prefer everything in one package it's just 5 files anyway.

finkmanAtSap:
Thank you for your review and the helpful remarks 😊

PR #1153, opened 2023-04-13
Fix ServiceBindingMapper for IAS configuration without domains

finkmanAtSap:
For backward compatibility, it also accepts single String values instead of String arrays for 'domains'. Single String values may also be provided via key 'domain'.


liga-oz (approved):
💯

finkmanAtSap:
> All in al looks very good, but the cast from string to string list for `domains` is not semantically correct, because domains shouldn't ever come from IAS binding as a string, therefore I would say you can remove that casting part.

Thanks for hint :) Makes the code much cleaner <3

PR #1172, opened 2023-04-19
Update samples/spring-security-basic-auth README

finkmanAtSap:


liga-oz (approved):
Overall really nice readme and thoughtful of you to finally update the screenshot 💯   I would just ask you before you merge it to decrease the size of the screenshot. It's very, very wide so on smaller screen the letters get squashed and are poorly visible, a good img size would be with aspect ratio 16:9 approx .1280 px by 720 px and highlight or circle in red the things we want the reader to pay attention to.

finkmanAtSap:
> highlight or circle in red the things we want the reader to pay attention to.

Sorry, I used green in the new screenshot because I didn't read properly. I hope it's okay, too :)

PR #1202, opened 2023-05-15
Fix flaky

akib6075:


akib6075 (commented):
compareObjects can not compare the different order of the properties of same two object. This change can resolve this flaky.

PR #1251, opened 2023-08-03
expose http status code in OAuth2ServiceException

Exotrom:
Expose the `httpStatusCode` field from `OAuth2ServiceException`.

It has been like this before, but the method had been removed in #1179.
If there is no reason not to expose it to the caller, we would like to have it back as our current code is relying on it for error handling.


liga-oz (approved):
thanks for the PR 💯 it looks like it was removed by mistake, it'll be available again in the next release, which is going out today.

PR #1276, opened 2023-08-28
Only create HttpClient bean if required

beckermarc:
The HttpClient bean is created for the XsuaaTokenFlows bean. However if the XsuaaTokenFlows bean is not created at all, the HttpClient still is. With this change the HttpClient is bound to the same conditions as the XsuaaTokenFlows bean.


beckermarc:
@liga-oz Could you please take a look here? This would avoid creating the unnecessary bean in CAP scenarios, where dependency `resourceserver-security-spring-boot-starter` is present, but no IAS or XSUAA binding has been configured.

liga-oz (approved):
thanks for this fix 💯

PR #1321, opened 2023-10-19
Minor Performance Improvement: Avoid Several Redundant Calls to Environment Layer During Bootstrapping

clnative:
During a debugging session, I stumbled over a lot of redundant calls to the `CFEnvironment` class originating from `IdentityServicesPropertySourceFactory`. Although not causing any external calls, still the performance may be degraded slightly due to this. 
It seems rather easy to prevent the unnecessary calls by introducing a local variable. IMO that also makes the code easier to read. The pattern may be applied both when determining XSUAA service instances and IAS service instances. I have adjusted both occurrences.

As this change is expected to be trivial (bugfix contribution only) and should be non-disruptive for any consumer, I don't expect any credit to be mentioned.

## Dependency
- [ ] requires #1314 or #1316 (whichever is chosen) to fix the pipeline/voter; no technical dependency known.

### DIFF
--- a/spring-security/src/main/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactory.java
+++ b/spring-security/src/main/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactory.java
@@ -102,13 +102,14 @@ public class IdentityServicesPropertySourceFactory implements PropertySourceFact
 	@Nonnull
 	private Properties getIasProperties(Environment environment) {
 		Properties properties = new Properties();
-		if (environment.getIasConfiguration() != null) {
+		final OAuth2ServiceConfiguration iasConfiguration = environment.getIasConfiguration();
+		if (iasConfiguration != null) {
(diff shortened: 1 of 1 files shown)


clnative:
Maven build for Java 8 failed with

```
[INFO] token-client ....................................... FAILURE [02:34 min]

Error:  One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0': 
Error:  
Error:  json-20230618.jar: CVE-2023-5072(7.5)
```

which looks unrelated to this >PR.

As a possible solution I see https://mvnrepository.com/artifact/org.json/json/20231013 to have been released, which apparently addresses this CVE.

clnative:
- [ ] depends on #1314 and #1316

finkmanAtSap:
> I don't think that this test error is related to the content of this PR.

Agreed. Once we have some time, we will take a look to see where the failed tests are coming from.

liga-oz (approved):

PR #1324, opened 2023-10-21
Make Selection of XSUAA/IAS Service Instances from Environments reproducible/predictable

clnative:
## Background

My (application dev) team and I currently face the situation that we have a BTP CF application, which is bound to (in total) three different XSUAA service instance.
It is already clarified that Spring Boot AutoConfiguration of `spring-security` will not do the trick. That is why we are trying to overwrite the configuration with our own.
We try to stay as close as possible to the infrastructure cloud-security-services-integration provides/uses to avoid incompatibilities. 

## Current Situation

During our debugging sessions we detected that the incapability of handling more than two XSUAA service instance is not just with AutoConfiguration, but goes deeper - down to the level of handling the `CFEnvironment`. The main problem we identified is located at https://github.com/clnative/cloud-security-services-integration-library/blob/6c6bbb7a38c60da62c4f19aa4985bf795e318c5f/env/src/main/java/com/sap/cloud/security/config/cf/CFEnvironment.java#L188. There an "arbitrary first instance" is selected. However, no sorting is applied before. That is why, depending in which order the service instances are mentioned in environment variable `VCAP_SERVICES` plus some other internal factors, "some" service instance is returned. This arbitrary selection is very nasty - in our case in our test landscapes all tests returned the expected service instance (by chance). As soon as we started to roll out the "validated implementation" to production, the wrong service instance was selected and promptly caused a downtime. 

## Solution Approach

Avoid "arbitrary" selection of the XSUAA/IAS service instances by taking the service instance with the lexicographically lowest CF service instance environment `name`[1]. This makes the selection behavior much more predictable and may even provide a chance for application developers to influence indirectly which service instance will be selected.

After alignment, we discussed to protect this sorting logic behind a configuration option to ensure backward compatibility. The configuration option gets explained in the `spring-security/README.md` file.

## Known Open Questions / Tasks

- [ ] Currently the approach is only thought through for the CF case. K8S cannot be tested by the author (atm). So far, no implementation for the K8S case has been done, as there is too little knowledge about how the situation is there. It needs to be clarified if the same approach shall also be applied there. At the current state of implementation, K8S does not consider any "service instance name" and therefore most likely would yield the old behavior.
- [x] Update of documentation required? Which, where?

## Footnotes

[1] Note that although typically this is the service instance's name (as used when creating the service instance), it may also be overwritten by a binding name. See also section "VCAP_SERVICES" in https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html.

### DIFF
--- a/env/src/main/java/com/sap/cloud/security/config/ServiceBindingMapper.java
+++ b/env/src/main/java/com/sap/cloud/security/config/ServiceBindingMapper.java
@@ -14,7 +14,7 @@ import static com.sap.cloud.security.config.Service.IAS;
 import static com.sap.cloud.security.config.cf.CFConstants.IAS.DOMAIN;
 import static com.sap.cloud.security.config.cf.CFConstants.IAS.DOMAINS;
 import static com.sap.cloud.security.config.cf.CFConstants.SERVICE_PLAN;
-import static com.sap.cloud.security.config.cf.CFConstants.NAME;;
+import static com.sap.cloud.security.config.cf.CFConstants.NAME;
 
 @Deprecated
 public class ServiceBindingMapper {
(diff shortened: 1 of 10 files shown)


clnative:
As long as this PR is in draft state, no success on the maven voters are expected.

clnative:
closed in favor of #1334

PR #1333, opened 2023-10-31
App2Service support

mwdb:
This is an initial draft for supporting App2Service flows.
What is missing:
1) In Spring applications the certificate is not available in the security context, see https://github.com/SAP/cloud-security-services-integration-library/issues/1332
2) ProofToken data is not passed back to the application. Tbd how data is passed from a validator to the application. Maybe extending the Token interface for this?
3) ProofToken data does not contain the plan. I guess this would be needed for AMS to be able to react on it.
4) [X509Certificate](https://github.com/SAP/cloud-security-services-integration-library/blob/main/java-security/src/main/java/com/sap/cloud/security/x509/X509Certificate.java#L67) is using X500Principal.RFC1779 formatting, while standard in identity broker is RFC2253.  Also the trim() looks a bit weird.


mwdb:
closing, as outdated approach

PR #1334, opened 2023-11-01
Enable IdentityServicesPropertySourceFactory to Map More Than Two XSUAA Service Instances

clnative:
In an offline discussion around #1324, we determined that the approach applied there is not really sustainable. It turned out that the root cause of the issue is more to be found at the limited implementation of `IdentityServicesPropertySourceFactory`, which does not pass through all Xsuaa Service Instances to class `XsuaaServiceConfigurations`. This effectively prohibits that consuming applications may overwrite AutoConfiguration and pick&choose the right XSUAA service instance they want to use when overwriting `JwtDecoder`. 

This PR tries to overcome this situation by properly mapping additional Xsuaa Service Instance configurations to the array fields after the reserved items 0 and 1. 

Note: This PR alone might not be sufficient to allow consumers properly determine which Xsuaa Service configuration they want to use, because they may still be unable to properly distinguish two service configurations from each other (by attribute values). 
However, this PR is still considered a necessary prerequisite for that as otherwise no access to the service instance's configuration during spring configuration is possible.

Closes #1324


clnative:
> DefaultHttpClientFactoryTest.closeHttpClient:97 expected: <200> but was: <403>

Doesn't look like an issue of this PR.

finkmanAtSap:
> > DefaultHttpClientFactoryTest.closeHttpClient:97 expected: <200> but was: <403>
> 
> Doesn't look like an issue of this PR.

There was an old unit test against an internet URL that introduced a Heisenbug when the server replied with 403 instead of 200. Should hopefully be resolved now after merging the fix into this.

finkmanAtSap (approved):
LGTM, nice work!

PR #1338, opened 2023-11-06
Fix for flaky test AuthenticationTokenTest.equals

219sansim:
**Overview**
The `AuthenticationTokenTest.Equals` fails because the token generated is not the same everytime. The `JSONObject.toString()` gives a non-deterministic output because because `JSONObject` values are unordered. The solution is to sort the key-value pairs of `jsonHeader` and `jsonPayload` before converting it to their string representation. I had to modify `XsuaaTokenCompTest.java`  as it was affected by this change and had a failing test.
**Command to reproduce the failure:**
```mvn -pl spring-security edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=com.sap.cloud.security.spring.token.authentication.AuthenticationTokenTest#equals```
**Error Output:**
The test fails on 3/5 iterations of the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool
```
 Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   AuthenticationTokenTest.equals:40 expected: <AuthenticationToken [Principal=Jwt header
        {"alg":"RS256","kid":"default-kid-ias"}
Jwt payload
        {"exp":6974056800,"aud":"theClientId","scim_id":"the-user-id","app_tid":"the-app-tid","zone_uuid":"the-zone-id","user_uuid":"the-user-id","azp":"theClientId","cid":"theClientId"}
, Credentials=[PROTECTED], Authenticated=true, Details=null, Granted Authorities=[]]> but was: <AuthenticationToken [Principal=Jwt header
        {"kid":"default-kid-ias","alg":"RS256"}
Jwt payload
        {"aud":"theClientId","zone_uuid":"the-zone-id","exp":6974056800,"azp":"theClientId","cid":"theClientId","app_tid":"the-app-tid","scim_id":"the-user-id","user_uuid":"the-user-id"}
, Credentials=[PROTECTED], Authenticated=true, Details=null, Granted Authorities=[]]>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```
**Reason:**
The `org.json.JSONObject` class uses Maps as the data structure for the `JSONObject`. And, according to [HashMap documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html),
_"This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time."_
This causes the `jsonHeader.toString()` and `jsonPayload.toString()` to give a non-deterministic output.
**About Flaky Tests:**
**Flaky tests** are tests in software development that produce inconsistent or unreliable results, which can lead to non-deterministic outcomes of the test case, fixing them is important to both reliable testing and fixing vulnerabilities in the code.


liga-oz:
follow up PR was merged https://github.com/SAP/cloud-security-services-integration-library/pull/1403

PR #1346, opened 2023-11-08
Update JwtBearerTokenFlow.java

borisovkalin:
Added additional setter for an opaque token responses.

### DIFF
--- a/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenServiceConstants.java
+++ b/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenServiceConstants.java
@@ -35,7 +35,7 @@ public class OAuth2TokenServiceConstants {
 	public static final String GRANT_TYPE_CLIENT_X509 = "client_x509";
 	public static final String GRANT_TYPE_AUTHORIZATION_CODE = "authorization_code"; // not supported by token-client
 																						// lib
-
+	public static final String TOKEN_FORMAT = "token_format";
 	public static final String TOKEN_TYPE_OPAQUE = "opaque";
 
 	public static final String PARAMETER_CLIENT_ID = "client_id";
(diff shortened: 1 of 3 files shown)


finkmanAtSap (changes requested):
Hi Kalin, thanks for your contribution.

Can you please make the following changes?

- change method to "setOpaqueTokenFormat(boolean opaque)" that allows both enabling or disabling the opaque format similiar to disableCache(boolean) that is already there.
- add a unit test here https://github.com/SAP/cloud-security-services-integration-library/blob/dd16f7513d3876fc6be0ad032ae9aef995f585f9/token-client/src/test/java/com/sap/cloud/security/xsuaa/tokenflows/JwtBearerTokenFlowTest.java#L129 for the new method that verifies that the optionalParameters contains the correct setting after using the method
- add a constant for "token_format" similar to "token_type" here: https://github.com/SAP/cloud-security-xsuaa-integration/blob/e5f4196a9f88d207b43664e421f88bbf56969d06/token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenServiceConstants.java#L17
- use the constant in your code and in addition in the two places where "token_format" has been hard-coded before: https://github.com/SAP/cloud-security-xsuaa-integration/blob/e5f4196a9f88d207b43664e421f88bbf56969d06/token-client/src/test/java/com/sap/cloud/security/xsuaa/client/XsuaaOAuth2TokenServiceJwtBearerTokenTest.java#L131
https://github.com/SAP/cloud-security-xsuaa-integration/blob/e5f4196a9f88d207b43664e421f88bbf56969d06/token-client/src/test/java/com/sap/cloud/security/xsuaa/client/XsuaaOAuth2TokenServicePasswordTest.java#L146

Thanks!
Manuel

finkmanAtSap (changes requested):
Thanks for your changes. Looks very good. Just one small thing. Then I would like to merge it:
Can you please add one more assertion to the unit test that by default (if setOpaqueTokenFormat was not called yet), it does not contain the token_format key = opaque entry?
And likewise, at the end of the method, add the same assertion after calling setOpaqueTokenFormat(false) after having called setOpaqueTokenFormat(true) before?

finkmanAtSap (approved):
Perfect, thank you for the adjustments!

PR #1354, opened 2023-11-21
Jwt Signature validator update

liga-oz:


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtSignatureValidator.java
```diff
@@ -54,12 +56,10 @@ private PublicKey fetchPublicKey(Token token, JwtSignatureAlgorithm algorithm) t
             throw new IllegalArgumentException("Token does not contain the mandatory " + KID_PARAMETER_NAME + " header.");
         }
 
-        String jwksUri = configuration.isLegacyMode() ? configuration.getUrl() + "/token_keys" : token.getHeaderParameterAsString(JKU_PARAMETER_NAME);
-        if (jwksUri == null) {
-            throw new IllegalArgumentException("Token does not contain the mandatory " + JKU_PARAMETER_NAME + " header.");
-        }
-
+        String jwksUri = configuration.isLegacyMode() ? configuration.getUrl() + "/token_keys" : configuration.getProperty(UAA_DOMAIN) + "/token_keys";
```
finkmanAtSap, line 59: Do we need a fallback to "url" property from configuration if UAA_DOMAIN is not available? 
We are supposed to add the fallback for XSA configuration support. Or would XSA configurations have isLegacyMode() return true anyways?

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java
```diff
@@ -127,8 +126,7 @@ private Jwt verifyToken(JWT jwt) {
 	private Jwt verifyToken(String token, String jku, String kid, String uaaDomain) {
 		try {
 			canVerifyWithKey(jku, kid, uaaDomain);
-			validateJku(jku, uaaDomain);
-			return verifyWithKey(token, jku, kid);
+			return verifyWithKey(token, composeJku(uaaDomain), kid);
```
finkmanAtSap, line 129: We need to pass the zid from the token to composeJku and there, append it as query parameter to the URL, e.g. .../token_keys?zid=\<tokenZid\>.
If it is not in the token or blank, this query parameter can be omitted.

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtSignatureValidator.java
```diff
@@ -54,12 +56,10 @@ private PublicKey fetchPublicKey(Token token, JwtSignatureAlgorithm algorithm) t
...
+        URI uri = URI.create(jwksUri);
+        uri =  uri.isAbsolute() ? uri : URI.create("https://" + jwksUri);
```
finkmanAtSap, line 61: We also need to put the zid from the token as query parameter into the URL, e.g. .../token_keys?zid=<tokenZid>.
If it is not in the token or blank, this query parameter can be omitted.

finkmanAtSap (changes requested):

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java
```diff
@@ -123,10 +125,26 @@ private Jwt verifyToken(JWT jwt) {
...
+		}
+		if (zid != null && zid.isBlank()){
+			zid = null;
+		}
+		return zid;
+	}
+
+	private Jwt verifyToken(String token, String jku, String kid, String uaaDomain, String zid) {
 		try {
 			canVerifyWithKey(jku, kid, uaaDomain);
```
finkmanAtSap, line 146: remove jku method parameter and use composeJku(uaaDomain, zid) instead? Then only pass jku and kid to canVerifyWithKey?

spring-xsuaa/src/main/java/com/sap/cloud/security/xsuaa/token/authentication/XsuaaJwtDecoder.java
```diff
@@ -111,7 +113,7 @@ private Jwt verifyToken(JWT jwt) {
 			String jku = tokenInfoExtractor.getJku(jwt);
 			String kid = tokenInfoExtractor.getKid(jwt);
 			String uaaDomain = tokenInfoExtractor.getUaaDomain(jwt);
-			return verifyToken(jwt.getParsedString(), jku, kid, uaaDomain);
+			return verifyToken(jwt.getParsedString(), jku, kid, uaaDomain, getZid(jwt));
```
finkmanAtSap, line 116: remove jku variable from line 113 and adjust verifyToken method signature to not take jku parameter anymore since it is a private method?

PR #1392, opened 2023-12-12
Use Jetty 12 with EE9 with Spring Boot 3.2

beckermarc:
Fixes https://github.com/SAP/cloud-security-services-integration-library/issues/1368


finkmanAtSap:
Do you know if the lib will stay backward-compatible with Spring Boot < 3.2 when we adopt the Jetty 12 namespaces?

I don't really like being dependent on Spring Boot to provide classes that our lib needs. I wonder if we can fix our dependencies if we depend on a specific Jetty version instead of relying on Spring Boot to provide the right version that fits our code.

liga-oz (approved):
Thanks a lot for your contribution 👍🏻

PR #1393, opened 2023-12-12
Include service plan in IdentityServicesPropertySourceFactory

finkmanAtSap:
When putting configurations into the Spring properties via IdentityServicesPropertySourceFactory, include the "plan" property that is already present on the OAuth2ServiceConfiguration objects used as input.


spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryMultiTest.java
clnative, line 1: I am unsure if I understand the motivation of removing this unit test. I agree that it is somewhat redundant to `IdentityServicesPropertySourceFactoryFourXsuaaOneIasTest`, but I have the impression that the use case tested is still valid.
Perhaps you have something more on your mind on this...

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryBrokerNoHoleTest.java
```diff
@@ -5,38 +5,68 @@
...
-		assertEquals("http://domain.xsuaadomain", configuration.xsuaaUrl0);
-		assertEquals("xsuaadomain", configuration.xsuaaDomain0);
-		assertEquals("xsappname2", configuration.xsuaaAppName0);
+		OAuth2ServiceConfiguration xsuaaConfig = env.getXsuaaConfiguration();
+		assertEquals(xsuaaConfig.getClientId(), configuration.xsuaaClientId0);
+		assertEquals(xsuaaConfig.getClientSecret(), configuration.xsuaaClientSecret0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.URL), configuration.xsuaaUrl0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.XSUAA.UAA_DOMAIN), configuration.xsuaaDomain0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.XSUAA.APP_ID), configuration.xsuaaAppName0);
+		assertEquals(ServiceConstants.Plan.APPLICATION, ServiceConstants.Plan.from(configuration.xsuaaPlan0));
```
clnative, line 61: With checking against `xsuaaConfig` your unit test does not longer test `IdentityServicesPropertySourceFactory` in isolation. Instead, you are also testing `env.getXsuaaConfiguration()` as well. This makes this test already something like a "(module) integration test".
I don't know the detailed guidelines in your project, that is why I am unsure if this is acceptable.

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryFourXsuaaOneIasTest.java
```diff
@@ -5,53 +5,86 @@
...
-		assertEquals("http://domain.xsuaadomain", configuration.xsuaaUrl0);
-		assertEquals("xsuaadomain", configuration.xsuaaDomain0);
-		assertEquals("xsappname2", configuration.xsuaaAppName0);
+		OAuth2ServiceConfiguration xsuaaConfig = env.getXsuaaConfiguration();
+		assertEquals(xsuaaConfig.getClientId(), configuration.xsuaaClientId0);
+		assertEquals(xsuaaConfig.getClientSecret(), configuration.xsuaaClientSecret0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.URL), configuration.xsuaaUrl0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.XSUAA.UAA_DOMAIN), configuration.xsuaaDomain0);
+		assertEquals(xsuaaConfig.getProperty(ServiceConstants.XSUAA.APP_ID), configuration.xsuaaAppName0);
+		assertEquals(ServiceConstants.Plan.APPLICATION, ServiceConstants.Plan.from(configuration.xsuaaPlan0));
```
clnative, line 61: (same as above)

spring-security/src/test/java/com/sap/cloud/security/spring/config/XsuaaServiceConfigurationsTest.java
```diff
@@ -5,30 +5,55 @@
...
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
 
+import java.io.IOException;
+import java.util.List;
+
+import static com.sap.cloud.security.spring.config.XsuaaServiceConfigurationTest.assertConfigsAreEqual;
```
clnative, line 23: Qualms... in `src/test/main/` you may have util classes, which will host such "helper methods".

spring-security/src/test/java/com/sap/cloud/security/spring/config/XsuaaServiceConfigurationTest.java
```diff
@@ -5,26 +5,54 @@
...
 	void configuresXsuaaServiceConfiguration() {
-		runner.withUserConfiguration(EnablePropertiesConfiguration.class)
-				.withPropertyValues("sap.security.services.xsuaa.url:http://localhost",
-						"sap.security.services.xsuaa.uaadomain:localhost", "sap.security.services.xsuaa.clientid:cid")
-				.run(context -> assertEquals("http://localhost",
-						context.getBean(XsuaaServiceConfiguration.class).getUrl().toString()));
+		assertConfigsAreEqual(xsuaaConfig, env.getXsuaaConfiguration());
+    }
+
+	static void assertConfigsAreEqual(XsuaaServiceConfiguration xsuaaConfig, OAuth2ServiceConfiguration oauthConfig) {
```
clnative, line 46: Method is no specific to this test and therefore should be extracted into a separate class (see also other review comment)

clnative (commented):
I like your changes a lot in general, although it's a mixture of introducing `plan` to `OAuth2ServiceConfiguration` via `IdentityServicesPropertySourceFactory` and a refactoring of the unit tests.

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryMultiTest.java
finkmanAtSap: Hm, not really. I just thought if it works correctly with 2 Application + 2 Broker, it will also work correctly with 1 Application + 1 Broker and wanted to reduce maintenance effort if the tests have to be adjusted again.
Let's see what Liga thinks.

clnative: okay, fine and agreed.

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryBrokerNoHoleTest.java
finkmanAtSap: Good point. My intention was to use this opportunity to test the backward-compatibility aspect that `IdentityServicesPropertySourceFactory` has to respect since it contains more than two configurations.

I see three options how to resolve it. Which one do you find best?

1. Changing the `IdentityServicesPropertySourceFactory` tests to assert **both** against the string values and against env.getXsuaaConfiguration()?

2. Accept the backward-compatibility aspect as sufficiently tested in `XsuaaServiceConfigurationsTest` where `IdentityServicesPropertySourceFactory` is used to populate the Spring properties. Than again, that test has the same problem you describe.

3. Add a dedicated (module) integration test and revert the `IdentityServicesPropertySourceFactory` tests and `XsuaaServiceConfiguration(s)Test` to test in isolation. Do you have a suggestion for a good name?

finkmanAtSap: @liga-oz I have option 3 committed locally in case we want to go with this. It was not that much work since we had the tests in the git history already with/without isolated testing by now. Let me know if I should push when you have time to review.

spring-security/src/test/java/com/sap/cloud/security/spring/config/ConfigurationUtil.java
```diff
@@ -0,0 +1,16 @@
+package com.sap.cloud.security.spring.config;
+
+import com.sap.cloud.security.config.OAuth2ServiceConfiguration;
+import com.sap.cloud.security.config.ServiceConstants;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ConfigurationUtil {
```
liga-oz, line 8: I would move this method it to the XsuaaServiceConfigurationTest class as it's the only place where it is used.

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryBrokerNoHoleTest.java
liga-oz: lets go with the option 3 it's my preferred option too and as you already done the work ,makes even more sense to go with it 😄

liga-oz (changes requested):
See comments.

spring-security/src/test/java/com/sap/cloud/security/spring/config/ConfigurationUtil.java
clnative: Well, not really - it has been our discussion before that it is also used here https://github.com/SAP/cloud-security-services-integration-library/pull/1393/commits/b93553697ac9b1cadae98cb4652da74f062483ea#diff-f1efe439764be3d82df836e42d26dd47e909646b0ea7c00a4fe8d375b4766de6
(XsuaaServiceConfigurationTest  and XsuaaServiceConfiguration**s**Test )

finkmanAtSap:
I just pushed the additional module integration tests between XsuaaServiceConfiguration(s) and IdentityServicesPropertySourceFactory (aka "option 3")

Also, I pushed a change to include the name of the service binding as property into the configuration.
I added tests to assert that this name property is also available inside the configuration beans of spring-security, so consumers can now filter the list of configurations based on the names of their service instances.
This was a missing feature that required consumers like @clnative to write hacky code using the suffix in xsappname to find specific configurations in the list when overriding the auto-configuration for their individual setup.

Can you please check again, if it looks good to you? Thanks!

spring-security/src/test/java/com/sap/cloud/security/spring/config/IdentityServicesPropertySourceFactoryMultiTest.java
finkmanAtSap: @liga-oz Is it okay if we remove this explicit 1 App + 1 Broker test because we added the 2 App + 2 Broker test?

liga-oz: that should be fine

liga-oz (approved):
Please rename the ConfigurationUtil class, but other than that looks good 👍🏻

env/src/main/java/com/sap/cloud/security/config/ServiceBindingMapper.java
```diff
@@ -40,8 +41,8 @@ public static OAuth2ServiceConfigurationBuilder mapToOAuth2ServiceConfigurationB
 		TypedMapView credentials = TypedMapView.ofCredentials(b);
 		OAuth2ServiceConfigurationBuilder builder = OAuth2ServiceConfigurationBuilder.forService(service)
 				.withProperties(credentials.getEntries(String.class))
-				.withProperty(SERVICE_PLAN,
-						b.getServicePlan().orElse(ServiceConstants.Plan.APPLICATION.name()).toUpperCase());
+				.withProperty(NAME,	b.getName().orElse(""))
+				.withProperty(SERVICE_PLAN,	b.getServicePlan().orElse(ServiceConstants.Plan.APPLICATION.name()).toUpperCase());
```
clnative, line 45: upsy, formatting looks a little strange here...

clnative (commented):
No objection, only one minor thing detected...

PR #1418, opened 2024-01-05
Use jetty-ee10 instead of jetty-ee9

beckermarc:
Follow up for #1392


beckermarc:
@liga-oz: Can you please have a look at this PR? This is a follow up of my PR were I migrated to Jetty 12. With this change the latest EE Version is now also used with Jetty 10, enabling Servlet 6 APIs (as is the case with Spring Boot 3)

liga-oz (approved):
lgtm 👍🏻

PR #1427, opened 2024-01-11
Remove uppercasing of service plan property when building OAuth2ServiceConfiguration from service bindings

finkmanAtSap:

PR #1429, opened 2024-01-12
Fix service plan casing and add binding name in ServiceBindingMapper

finkmanAtSap:
Combines #1427 and #1393

### DIFF
--- a/env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
+++ b/env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
@@ -59,7 +59,11 @@ public class K8sEnvironment implements Environment {
 				.map(OAuth2ServiceConfigurationBuilder::build)
 				.collect(Collectors.toMap(config -> Plan.from(config.getProperty(SERVICE_PLAN)),
 						config -> config,
-						(l,r) -> {throw new IllegalArgumentException("2 service configurations of the same plan are not supported");},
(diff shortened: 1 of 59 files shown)


env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
```diff
@@ -58,14 +58,14 @@ private void loadAll() {
 				.map(ServiceBindingMapper::mapToOAuth2ServiceConfigurationBuilder)
 				.filter(Objects::nonNull)
 				.map(OAuth2ServiceConfigurationBuilder::build)
-				.collect(Collectors.toMap(config -> config.getProperty(SERVICE_PLAN),
+				.collect(Collectors.toMap(config -> config.getProperty(SERVICE_PLAN) != null ? config.getProperty(SERVICE_PLAN).toUpperCase() : "",
 						Function.identity()));
```
liga-oz, line 62: why here and below is toUpperCase() used? Is it really necessary?

liga-oz (approved):
Overall looks good just one remark about the upper case in the K8sEnvironment class

env/src/main/java/com/sap/cloud/security/config/k8s/K8sEnvironment.java
liga-oz: sounds good and you can use than `EnumMap`

PR #1453, opened 2024-02-07
Validate uaadomain earlier in XsuaaJwtDecoder to prevent NPE

finkmanAtSap:
Bugfix for a reported NPE during composeJku when accessing uaadomain which is null.
I simply moved the validation of the JWKS parameters (kid, uaadomain) to an earlier place which, instead of running into the NPE, should immediately cause the fallback to the verificationkey from the credentials.


liga-oz (approved):
💯

PR #1461, opened 2024-02-15
Fix missing exception

MarievonSeggern:
Fixed security issue "Missing catch of NumberFormatException" in parseJwt method


spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
```diff
@@ -59,10 +59,15 @@ public Mono<Jwt> decode(String encodedToken) throws JwtException {
...
-                token.getExpiration(), token.getHeaders(), token.getClaims()));
+        try{
+            Instant issuedAt = token.hasClaim(TokenClaims.XSUAA.ISSUED_AT)
+                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaimAsString(TokenClaims.XSUAA.ISSUED_AT)))
+                    : null;
+            return Mono.just(new Jwt(token.getTokenValue(), issuedAt,
+                    token.getExpiration(), token.getHeaders(), token.getClaims()));
+        }
+        catch (NumberFormatException e){
+            throw new RuntimeException("Error parsing JWT: " + e.getMessage(), e);
```
liga-oz, line 70: I would prefer here to use BadJwtException instead of the generic runtime exception

liga-oz (changes requested):
Please change the RuntimeException with a more dedicated one.

liga-oz (approved):
💯

PR #1474, opened 2024-03-05
Fix for usage of WebServlet annotation in java-security-test

MarievonSeggern:
In the sample libraries that used the @WebServlet annotation the tests didn't work. 
I made changes in the java-security-test library so that the base resource path of WebAppContext is now set dynamically based on the existence of a web.xml file.


liga-oz (commented):
How come that previously the pipeline worked when building the java-security-usage-ias sample without these changes? Are you sure that this change fixes some issue? Or have you not committed all changes in this PR?

PR #1492, opened 2024-03-27
Extend SecurityContext with the service broker plan name for app2service communications 

liga-oz:
use cases with proofToken check

### DIFF
--- a/java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
+++ b/java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
@@ -6,6 +6,8 @@
 package com.sap.cloud.security.token;
 
 import com.sap.cloud.security.x509.Certificate;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -22,7 +24,7 @@ public class SecurityContext {
 	}
 
 	private static final ThreadLocal<Token> tokenStorage = new ThreadLocal<>();
-	private static final ThreadLocal<String[]> planStorage = new ThreadLocal<>();
+	private static final ThreadLocal<List<String>> servicePlanStorage = new ThreadLocal<List<String>>();
 	private static final ThreadLocal<Certificate> certificateStorage = new ThreadLocal<>();
 
 	/**
@@ -70,7 +72,6 @@ public class SecurityContext {
 		tokenStorage.set(token);
 	}
 
-
 	/**
 	 * Returns the token that is saved in thread wide storage.
 	 *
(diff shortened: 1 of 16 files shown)


token-client/src/test/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyServiceTest.java
```diff
@@ -118,6 +123,22 @@ public void retrieveTokenKeys_errorOccurs_throwsServiceException() throws IOExce
 				.hasMessageContaining(errorMessage);
 	}
 
+	@Test
+	public void retrieveTokenKeys_app2Service_ProofToken() throws IOException {
+		CloseableHttpResponse response = HttpClientTestFactory.createHttpResponseWithHeaders(jsonWebKeysAsString,
+				new BasicHeader[] { new BasicHeader(X_OSB_PLAN, "plan") });
```
finkmanAtSap, line 129: Let's try to get a better example from colleagues with the actual comma-separated list format

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyServiceTest.java
```diff
@@ -118,6 +123,22 @@ public void retrieveTokenKeys_errorOccurs_throwsServiceException() throws IOExce
...
+
+		when(httpClient.execute(any(), any(ResponseHandler.class))).thenAnswer(invocation -> {
+			ResponseHandler responseHandler = invocation.getArgument(1);
+			return responseHandler.handleResponse(response);
+		});
+
+		Map<String, String> requestParams = new HashMap<>();
+		requestParams.put(HttpHeaders.X_CLIENT_CERT, "cert");
+		cut.retrieveTokenKeys(TOKEN_KEYS_ENDPOINT_URI, requestParams);
+		assertNotNull(SecurityContext.getServicePlan());
```
finkmanAtSap, line 139: Let's change it to an assertion on the expected value

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -21,6 +21,7 @@ private SecurityContext() {
 	}
 
 	private static final ThreadLocal<Token> tokenStorage = new ThreadLocal<>();
+	private static final ThreadLocal<String> planStorage = new ThreadLocal<>();
```
finkmanAtSap, line 24: As the header is now supposed to contain a list of service plans separated by comma including escaped quotes (to be confirmed), maybe we should instead use ThreadLocal<List<String>> and convert the raw String header value to a list of plans without escape quote chars inside the setter? Otherwise parsing will be required on each access to the plans and there is no documentation yet which format consumers can expect this String to be.

finkmanAtSap (changes requested):
Your solution is much more minimal than I expected :) I like the approach and think it is the best way to implement the feature request if the server API will indeed supply the plan as a side-effect of an unrelated HTTP request inside the response header...

Given this server API, a "cleaner" solution inside the client lib where fetching the plan is not a (rather unexpected) side effect of signature validation, would be very verbose and require lots of refactoring to implement that would not be worth it in my opinion.

I think we should add some more documentation to DefaultOAuth2TokenKeyService to explain why the hell a service plan header is getting extracted when fetching the JWKS, so future maintainers will have a chance to understand this decision. The two features will be married inside the code now and anyone trying to refactor JWKS fetching in the future will have to know about this quirk.

Also, I think we should push colleagues to get actual response examples in the real format and consider making the getter return a parsed list of plans (see my comments).

Some final questions:

- I wonder if SpringOAuth2TokenKeyService also needs to have this header extraction implemented but I could not find any usage of the class outside of unit tests. Is it fine to not implement header extraction there?
- Do you think it is a problem that consumer who choose to use their own OAuth2TokenKeyService implementation via JwtValidatorBuilder will have to re-implement plan header extraction?

liga-oz:
> I wonder if SpringOAuth2TokenKeyService also needs to have this header extraction implemented but I could not find any usage of the class outside of unit tests. Is it fine to not implement header extraction there?

I couldn't find any usage either and it's intended to be used for Spring apps, but as our spring-security uses internally `DefaultOAuth2TokenKeyService` implementation I think we're good to go. I'm gonna mark that class as Deprecated and for removal, don't see a need for it any longer.

> Do you think it is a problem that consumer who choose to use their own OAuth2TokenKeyService implementation via JwtValidatorBuilder will have to re-implement plan header extraction?

If they provide their own implementation they must know what they do, so I guess it is completely fine. I Besides, I don't think there's that many out there who have done that.

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyService.java
```diff
@@ -74,8 +74,11 @@ public String retrieveTokenKeys(@Nonnull URI tokenKeysEndpointUri, Map<String, S
 				}
 
 				LOGGER.debug("Successfully retrieved token keys from {} with params {}.", tokenKeysEndpointUri, params);
+				// This is required for app2service communication. In case x-client_cert is provided in the request,
+				// the response can contain identity service broker plan header. Values
 				if (response.containsHeader(X_OSB_PLAN)) {
-					SecurityContext.setServicePlan(response.getFirstHeader(X_OSB_PLAN).getValue());
+					SecurityContext.setServicePlan(
+							response.getFirstHeader(X_OSB_PLAN).getValue().trim().split("\\s*,\\s*"));
```
finkmanAtSap, line 81: Maybe it would be better to move the parsing to setServicePlan, so the service plan logic is mostly there and the parsing does not have to be re-implemented if this OAuth2TokenKeyService is overridden?

finkmanAtSap (commented):
I'm not sure if the parsing by comma is enough. I remember Stefan mentioned something about escaped quote characters around the plan names.

finkmanAtSap (approved):

PR #1501, opened 2024-04-09
Fix for IAT Claim in ReactiveHybridJwtDecoder

### LINKED ISSUE #1490: ReactiveHybridJwtDecoder Tries to read Numeric IAT Claim as String
In the most recent `3.3.5` version of this library, the following code fails to allow proper handling of well-formed JWT tokens that include an `iat` claim:

https://github.com/SAP/cloud-security-services-integration-library/blame/3.3.5/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java#L69

This claim is, by definition, encoded as a `NumericDate` in the corresponding [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519):

> [4.1.6](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6).  "iat" (Issued At) Claim
The "iat" (issued at) claim identifies the time at which the JWT was
issued.  This claim can be used to determine the age of the JWT.  Its
value MUST be a number containing a **NumericDate** value.  Use of this
claim is OPTIONAL.

This is problematic due to the inclusion of this `iat` claim in JWT tokens generated in an XSUAA context. The issue itself manifests as an authentication error which has the corresponding logs:

```
2024-03-26T13:27:10.166-05:00 DEBUG 561231 --- [or-http-epoll-2] o.s.s.w.s.a.AuthenticationWebFilter:
  Authentication failed: Error initializing JWT decoder: JSONObject["iat"] is not a string (class java.lang.Integer : 1711477629).
```

This error can be replicated by deploying the sample application [spring-webflux-security-hybrid-usage](https://github.com/SAP/cloud-security-services-integration-library/tree/d47dc18c2fec83c8d0b90c011c97f70f5395ce70/samples/spring-webflux-security-hybrid-usage). One slight modification is needed to make the Spring application runnable which is changing the `@EnableWebSecurity` annotation in [`SecurityConfiguration.java`](https://github.com/SAP/cloud-security-services-integration-library/blob/d47dc18c2fec83c8d0b90c011c97f70f5395ce70/samples/spring-webflux-security-hybrid-usage/src/main/java/sample/spring/webflux/hybrid/SecurityConfiguration.java#L34) to `@EnableWebFluxSecurity`. 

After running the application and attempting to access the `/v1/sayHello` endpoint through the approuter and completeing the authentication, you will receive a 401 authentication error while the application records the log listed above.

Examining the callstack, we can see that when the token is initially parsed in `JSONObject`, it is done so as an `Integer`:
image

> Note: This is with the `JSONObject` present in the most recent version of `org.json:json` which is `20240205`

Given that this underlying value is stored as an `Integer` in the `JSONObject` the call to `token.getClaimAsString(TokenClaims.XSUAA.ISSUED_AT)` will fail.

MarievonSeggern:
IAT Claim (IssuedAt Claim) in XSUAA Token is numeric but was tried to read as a String. Fixes #1490

### DIFF
--- a/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
+++ b/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
@@ -66,7 +66,7 @@ public class ReactiveHybridJwtDecoder implements ReactiveJwtDecoder {
     static Mono<Jwt> parseJwt(Token token) {
         try{
             Instant issuedAt = token.hasClaim(TokenClaims.XSUAA.ISSUED_AT)
-                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaimAsString(TokenClaims.XSUAA.ISSUED_AT)))
+                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaims().get("iat").toString()))
                     : null;
             return Mono.just(new Jwt(token.getTokenValue(), issuedAt,
                     token.getExpiration(), token.getHeaders(), token.getClaims()));
--- a/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
+++ b/spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
@@ -66,7 +66,7 @@ public class ReactiveHybridJwtDecoder implements ReactiveJwtDecoder {
     static Mono<Jwt> parseJwt(Token token) {
         try{
             Instant issuedAt = token.hasClaim(TokenClaims.XSUAA.ISSUED_AT)
-                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaims().get("iat").toString()))
+                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaims().get(TokenClaims.XSUAA.ISSUED_AT).toString()))
                     : null;
             return Mono.just(new Jwt(token.getTokenValue(), issuedAt,
                     token.getExpiration(), token.getHeaders(), token.getClaims()));
--- a/spring-security/src/test/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoderTest.java
+++ b/spring-security/src/test/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoderTest.java
@@ -52,6 +52,7 @@ class ReactiveHybridJwtDecoderTest {
 
 		assertEquals("theClientId", cut.decode(encodedToken).block().getClaim(TokenClaims.AUTHORIZATION_PARTY));
         assertTrue(jwtToken.hasClaim(TokenClaims.XSUAA.ISSUED_AT));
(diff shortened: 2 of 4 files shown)


liga-oz (changes requested):
Please cover this in tests so that you're sure that this change really fixes the issue.

spring-security/src/test/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoderTest.java
```diff
@@ -46,6 +46,18 @@ void parseJwt() {
...
 
+	@Test
+	void parseJwtXsuaaToken() {
+		JwtGenerator jwtGeneratorXsuaa = JwtGenerator.getInstance(XSUAA, "theClientId").withClaimValue("iat", "1704063600");
+
+		Jwt jwtToken = ReactiveHybridJwtDecoder.parseJwt(jwtGeneratorXsuaa.createToken()).block();
+
+		assertEquals(3, jwtToken.getHeaders().size());
+		assertEquals(7, jwtToken.getClaims().size());
+		assertEquals(1, jwtToken.getExpiresAt().compareTo(Instant.now()));
```
liga-oz, line 57: I guess you want to check here the issuedAt (iat) not the expiresAt claim (ext) 🤔

liga-oz (changes requested):
I would highly recommend to run the newly added test before you applied the fix and after to make sure that it really covers the bug case.

spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoder.java
```diff
@@ -66,7 +66,7 @@ public Mono<Jwt> decode(String encodedToken) throws JwtException {
     static Mono<Jwt> parseJwt(Token token) {
         try{
             Instant issuedAt = token.hasClaim(TokenClaims.XSUAA.ISSUED_AT)
-                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaimAsString(TokenClaims.XSUAA.ISSUED_AT)))
+                    ? Instant.ofEpochSecond(Long.parseLong(token.getClaims().get("iat").toString()))
```
liga-oz, line 69: Why are you not using the Constant here?

spring-security/src/test/java/com/sap/cloud/security/spring/token/authentication/ReactiveHybridJwtDecoderTest.java
```diff
@@ -46,6 +46,13 @@ void parseJwt() {
 
 	}
 
+	@Test
+	void decodeXsuaaTokenWithIatClaim() {
+		String encodedToken = JwtGenerator.getInstance(XSUAA, "theClientId").withClaimValue("iat", "1704067200").createToken().getTokenValue();
+
+		assertEquals("theClientId", cut.decode(encodedToken).block().getClaim(TokenClaims.AUTHORIZATION_PARTY));
```
liga-oz, line 53: in this test where/how do you make sure that this test covers the bug fix?

liga-oz (changes requested):

liga-oz (approved):

PR #1511, opened 2024-04-12
Add toggle for the ProofToken validation

liga-oz:


java-security/README.md
```diff
@@ -271,22 +271,42 @@ This validator is not part of the default `CombiningValidator`, it needs to be a
...
+
+```java
+JwtX5tValidator validator=new JwtX5tValidator(oAuth2ServiceConfiguration);
+		ValidationResult result=validator.validate(token);
 ```
-Or it can be used as a standalone `Validator`, by creating a new instance of it and calling `JwtX5tValidator.validate(Token token)` method with the token to be validated as a method's parameter. See [here](#retrieve-additional-information-from-token) how to get a token from `SecurityContext`
+
+#### Proof Token validation
+
+Once enabled, it will forward the X509 client certificate from the request header `x-fowarded-client-cert`
```
finkmanAtSap, line 289: Once available, let us add a link to a documentation that describes this feature in detail. This description here is rather technical but for now, it is the best we can do.

finkmanAtSap (approved):

PR #1526, opened 2024-04-30
Improve the custom SecurityContext Strategy registration

beckermarc:
The library now registers a security context strategy by default. In CAP Java we ran into situations where this registration happened too late. The effect of this was, that SecurityConfigs used an obsolete SecurityContextStrategy (which stored by Spring Security as a member variable in the SecurityConfig) and the library set a new strategy in the static holder class. As the SecurityConfigs stored the authentication objects in the obsolete strategy, CAP wasn't able anymore to get the current authentication from the static holder class.

This PR tries to avoid this situation, by ensuring that the auto-configuration has a high precedence and is run early in the application initialization. Additional this PR avoids reflection usage to set the strategy, as reflection is not really required.

### DIFF
--- a/spring-security/src/main/java/com/sap/cloud/security/spring/autoconfig/SecurityContextAutoConfiguration.java
+++ b/spring-security/src/main/java/com/sap/cloud/security/spring/autoconfig/SecurityContextAutoConfiguration.java
@@ -5,14 +5,17 @@
  */
 package com.sap.cloud.security.spring.autoconfig;
 
-import org.springframework.beans.factory.config.MethodInvokingFactoryBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.boot.autoconfigure.AutoConfigureOrder;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
-import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


beckermarc:
@liga-oz Please have a look at this PR, thanks! It is currently breaking some scenarios of CAP Java when using XSUAA 3.4.2 with the now by default enabled security context strategy.

beckermarc:
Do not merge yet, still found some scenarios where this way of initialization creates problems.

beckermarc:
I found an approach that now works in all scenarios and makes sure the strategy is registered early enough. It is a little ugly, as it integrates with the web servlet initialization, which is slightly unrelated (although not totally).

Alternatively I also made another proposal to set the strategy based on an environment post processor. However this is slightly "incompatible" as it changes the way in which the configuration can be disabled (only properties, no auto config exclusion anymore): https://github.com/beckermarc/cloud-security-services-integration-library/pull/1

@liga-oz Let me know which approach you are preferring or if you see any problems with any of these suggestions.

liga-oz (approved):
🥇 thanks a million for the fix

beckermarc:
I actually noticed that this approach apparently seems to fail in MockMvc scenarios and runs too late there. This means that in those cases the registration happens too late again, breaking CAPs security configuration. Unfortunately I didn't seem to that that scenario thoroughly.

I reopened my alternative approach, that really makes sure to run this as early as possible, with the above mentioned limitations: No auto config exclusion any more, but only disabling through properties possible.

https://github.com/SAP/cloud-security-services-integration-library/pull/1536

@liga-oz Can you take another look at this please?

PR #1528, opened 2024-05-03
Support PKCS#8 standard keys from Zero trust Identity Service

liga-oz:
Sample App for testing ZTIS with XSUAA https://github.com/SAP/cloud-security-services-integration-library/tree/zero_trust_sample_app_xsuaa/samples/spring-security-hybrid-usage


finkmanAtSap (approved):
Nice work!

java-api/src/main/java/com/sap/cloud/security/config/ClientIdentity.java
```diff
@@ -67,6 +70,20 @@ default String getKey() {
 		return null;
 	}
 
+	/**
+	 * @return Certificate chain array
+	 */
+	default Certificate[] getCertificateChain() {
```
finkmanAtSap, line 76: should be @Nullable like the remaining getters?

java-api/src/main/java/com/sap/cloud/security/config/ClientIdentity.java
```diff
@@ -67,6 +70,20 @@ default String getKey() {
...
+		return null;
+	}
+
+	/**
+	 * @return Private key
+	 */
+	default PrivateKey getPrivateKey() {
```
finkmanAtSap, line 83: should be @Nullable like the remaining getters?

java-api/src/main/java/com/sap/cloud/security/config/ClientIdentity.java
```diff
@@ -34,7 +36,8 @@ default boolean isValid() {
 	 * @return the boolean
 	 */
 	default boolean isCertificateBased() {
-		return hasValue(getCertificate()) && hasValue(getKey());
+		return hasValue(getCertificate()) && hasValue(
```
finkmanAtSap, line 39: the formatting looks weird, can we put the two different combinations on one line each maybe?

token-client/src/main/java/com/sap/cloud/security/mtls/SSLContextFactory.java
```diff
@@ -91,8 +91,17 @@ public SSLContext create(String x509Certificates, String privateKey)
 	 */
 	public SSLContext create(ClientIdentity clientIdentity) throws GeneralSecurityException, IOException {
 		assertNotNull(clientIdentity, "clientIdentity must not be null");
-		assertHasText(clientIdentity.getCertificate(), "clientIdentity.getCertificate() must not return null");
-		assertHasText(clientIdentity.getKey(), "clientIdentity.getKey() must not return null");
+		try {
```
finkmanAtSap, line 94: I think this is try/catch is redundant because the same checks are performed inside createKeyStore which is the first statement following these lines

token-client/src/main/java/com/sap/cloud/security/mtls/SSLContextFactory.java
```diff
@@ -133,30 +142,52 @@ private SSLContext createDefaultSSLContext() throws NoSuchAlgorithmException {
...
-			keySpec = new PKCS8EncodedKeySpec(Base64.getDecoder().decode(removeHeaders(privateKey)));
+		if (privateKey != null && privateKey.getAlgorithm().equals("EC")) {
 			keyFactory = KeyFactory.getInstance("EC");
+			keySpec = new PKCS8EncodedKeySpec(privateKey.getEncoded());
 		}
-		if (logger.isDebugEnabled()) {
-			logger.debug("privateKeyPem: '{}...{}'", privateKey.substring(5, 40),
-					privateKey.substring(privateKey.length() - 25));
+
+		if (pemPrivateKey != null) {
```
finkmanAtSap, line 178: Is it intentionally not an *else if*?

My suggestion would be:
* remove both try/catch blocks from parent functions and inside _getPrivateKey_:
   * Call getPrivateKey() and getCertificateChain().
   * If both not null, use the PrivateKey instance directly if possible or create PrivateKey instance from it as done in line 175.
   * Else build PrivateKey from getKey() if getKey() and getCertificate() both not null/empty
   * Return PrivateKey if it was built in previous lines or throw the exception if not

finkmanAtSap (changes requested):
I think I understand the intention of the PR now and you can consider it approved. I just added some minor comments that you may or may not want to implement.

finkmanAtSap (approved):
Nice!

I'm happy with the changes but I have one more thought: We could have also moved the new logic into the ClientCertificate class which has one constructor for String-based certificate+key and one constructor for a PrivateKey/Certificate[] combination. Internally, the class could compute the missing parameters from the provided ones, such that the getters are all non-null, no matter what the instance was constructed from.
In the SSLContextFactory, one could then always call getPrivateKey and getCertificateChain without having to differentiate between the formats.
Not sure which alternative is better but in the second option, I think it is nicer that the support of different certificate formats/classes would be isolated in the ClientCertificate class and not spread across modules.

liga-oz:
> Nice!
> 
> I'm happy with the changes but I have one more thought: We could have also moved the new logic into the ClientCertificate class which has one constructor for String-based certificate+key and one constructor for a PrivateKey/Certificate[] combination. Internally, the class could compute the missing parameters from the provided ones, such that the getters are all non-null, no matter what the instance was constructed from. In the SSLContextFactory, one could then always call getPrivateKey and getCertificateChain without having to differentiate between the formats. Not sure which alternative is better but in the second option, I think it is nicer that the support of different certificate formats/classes would be isolated in the ClientCertificate class and not spread across modules.

the idea is quite good, but that would be quite a big chunk of logic that goes into certificate class and it would be a bit bigger refactoring and I would say too big for this PR. I'll create a BLI for this as it's not that high Prio but good to have thing 👍🏻

PR #1536, opened 2024-05-13
Set SecurityContextStrategy based on an EnvironmentPostProcessor

beckermarc:
Alternative proposal to https://github.com/SAP/cloud-security-services-integration-library/pull/1526 More explicit, by leveraging an EnvironmentPostProcessor, which runs very early in the startup phase and doesn't require the dependency to Servlet stuff introduced in the other PR. However looses the ability to exclude this AutoConfiguration through annotations or by defining a custom SecurityContextStrategy bean. Disabling the behaviour is only possible through properties with this approach.

Addon: Fixes an issue with MockMvc, as in this scenario the servlet initialization is not happening and the code runs too late due to that.


liga-oz (approved):
🥇

PR #1538, opened 2024-05-14
Support validation of user_token GrantType tokens

finkmanAtSap:
Add user_token again to GrantType enum to prevent errors in applications validating tokens granted by this grant_type.


liga-oz (approved):
💯

PR #1560, opened 2024-06-07
improve xsuaa jwk fetch error handling

liga-oz:


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/XsuaaJwtSignatureValidator.java
```diff
@@ -59,20 +59,27 @@ protected PublicKey getPublicKey(Token token, JwtSignatureAlgorithm algorithm)
...
+			if (configuration.hasProperty(ServiceConstants.XSUAA.VERIFICATION_KEY)) {
+				String fallbackKey = configuration.getProperty(ServiceConstants.XSUAA.VERIFICATION_KEY);
+				try {
+					key = JsonWebKeyImpl.createPublicKeyFromPemEncodedPublicKey(JwtSignatureAlgorithm.RS256,
+							fallbackKey);
+				} catch (NoSuchAlgorithmException | InvalidKeySpecException ex) {
+					IllegalArgumentException illegalArgEx = new IllegalArgumentException(
+							"Fallback validation key supplied via " + ServiceConstants.XSUAA.VERIFICATION_KEY
+									+ " property in service credentials could not be used: " + ex.getMessage());
+					if (e.getClass().equals(OAuth2ServiceException.class)) {
```
finkmanAtSap, line 76: Can the instanceof operator be used instead here?

spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/HybridJwtDecoder.java
```diff
@@ -82,7 +83,11 @@ public Jwt decode(String encodedToken) {
 		default -> throw new BadJwtException("Tokens issued by " + token.getService() + " service aren't supported.");
 		}
 		if (validationResult.isErroneous()) {
-			throw new BadJwtException("The token is invalid: " + validationResult.getErrorDescription());
+			if (validationResult.getErrorDescription().contains("Error retrieving token keys")) {
```
finkmanAtSap, line 86: We should probably check for [this](https://github.com/SAP/cloud-security-services-integration-library/blob/cf45ae45bb76e54b09aa315de7d3b07fbdcbbd4d/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureValidator.java#L48) error message instead because it is already doing a check for JWKS fetch failure there and it results in this higher level error message

finkmanAtSap (approved):
Just some minor thing, not that important

spring-security/src/main/java/com/sap/cloud/security/spring/token/authentication/HybridJwtDecoder.java
```diff
@@ -82,7 +83,11 @@ public Jwt decode(String encodedToken) {
...
+				throw new JwtException("Error retrieving token keys: " + validationResult.getErrorDescription());
```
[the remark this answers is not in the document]
liga-oz: good catch, that's not that DRY 🤣

PR #1565, opened 2024-06-12
Skip prooftoken check for app2app communications

liga-oz:


java-security/src/test/java/com/sap/cloud/security/token/validation/validators/SapIdJwtSignatureValidatorTest.java
```diff
@@ -170,6 +170,21 @@ public void validate_withEnabledProofTokenCheck() throws IOException {
 		assertTrue(cut.validate(iasToken).isValid());
 	}
 
+	@Test
+	public void validate_withEnabledProofTokenCheck_app2app() {
```
finkmanAtSap, line 174: Does this test really test the new functionality? I think we must check if tokenKeyServiceMock#retrieveTokenKeys has been called with or without certificate.

finkmanAtSap (changes requested):
Looks good, nice and simple! But I think the unit test should be slightly extended

PR #1566, opened 2024-06-12
Extend autoconfigurations for hybridJwtDecoders prooftoken check

liga-oz:


finkmanAtSap (changes requested):
I think this class is getting out of hand with too much corner cases. With every new flag, it will just multiply in complexity.
If I remember correctly, you said the extra bean definitions are necessary to unit test.
How about adding a getter to IasJwtDecoder that lets us check in the unit test whether its validator(s) have proof token enabled or not?

finkmanAtSap (approved):

PR #1572, opened 2024-06-18
ProofToken Autoconfiguration extension option 2

liga-oz:


liga-oz:
Closing as this was not the best solution

PR #1677, opened 2024-12-11
Adding support for envoy x-forwarded-client-cert header format

kuntzed:
see https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert


kuntzed:
Forgot to replace cert in tests with faked one.

PR #1679, opened 2024-12-11
Adding support for envoy x-forwarded-client-cert header format

kuntzed:
If parsing the x-forwarded-client-cert header as PEM fails, if possible the header is parsed based on the specification at: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert


SAP-MartinNaumann (approved):

finkmanAtSap (approved):
Approving on behalf of Martin

PR #1719, opened 2025-02-24
SCICAS-3386 Extend token validation results with retryability flag

kuntzed:
Validation results can now be flagged as retryable which is used to mark error while fetching JWKS. This allows clients to retry the validation when it may only have failed due to temporary network hickups while contacting the IAS tenant.

Jira: SCICAS-3386


NiklasHerrmann21:
Closed PR as its replaced with another implementation 
See: https://github.com/SAP/cloud-security-services-integration-library/pull/1742

PR #1722, opened 2025-02-28
add domains to equals/hashcode of OAuth2ServiceConfigurationImpl

finkmanAtSap:


kuntzed (approved):

marcel-hoell:
This leads to this issue: https://github.com/SAP/cloud-security-services-integration-library/issues/1743
This is a breaking change for us (and I think many others), unless other SAP libraries were updated also.

PR #1742, opened 2025-03-18
SCICAS-3386 Add retry to token key fetch 

NiklasHerrmann21:


token-client/src/main/java/com/sap/cloud/security/client/TokenClientConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param retryDelayTime The delay between retries.
+     */
+    void setRetryDelayTime(long retryDelayTime);
+
+    /**
+     * Gets the retry status codes.
+     *
+     * @return A comma-separated string of HTTP status codes.
+     */
+    String getRetryStatusCodes();
```
kuntzed, line 56: I would prefer this to return a `Collection<int>` or `Collection<HttpStatus>.` What do you think?

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenServiceConstants.java
```diff
@@ -7,36 +7,35 @@
...
+    public static final String REFRESH_TOKEN = "refresh_token";
+    public static final String TOKEN_TYPE = "token_type";
+    public static final String CLIENT_ID = "client_id";
+    public static final String CLIENT_SECRET = "client_secret";
+    public static final String USERNAME = "username";
+    public static final String PASSWORD = "password"; // NOSONAR
+    public static final String ASSERTION = "assertion";
+    public static final String AUTHORITIES = "authorities";
+    public static final String SCOPE = "scope";
+    public static final long RETRY_MAX_DELAY_TIME = 10000L;
```
kuntzed, line 21: Why is this required? And we would then also need a MAX_MAX_RETRY_ATTEMPTS, don't we?

kuntzed (changes requested):
This is a bit hard to review due to the many code reformattings. We should discuss if the reformatting is needed and if you want to keep it I suggest to first reformat the code in a separate change.

token-client/src/main/java/com/sap/cloud/security/xsuaa/client/OAuth2TokenServiceConstants.java
kuntzed, line 1: no real change in this file

token-client/src/test/java/com/sap/cloud/security/client/DefaultTokenClientConfigurationTest.java
```diff
@@ -0,0 +1,121 @@
...
+  @Test
+  public void setRetryStatusCodes_withEmptyString() {
+    config.setRetryStatusCodes("");
+    assertThat(config.getRetryStatusCodes()).isEmpty();
+  }
+
+  @Test
+  public void setRetryStatusCodes_withInvalidString() {
+    config.setRetryStatusCodes("abc,123");
+    assertThat(config.getRetryStatusCodes()).containsExactlyInAnyOrder(123);
```
kuntzed, line 72: not sure if we should really silently ignore malformed entries. did you think about failing in this case?

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyServiceTest.java
```diff
@@ -1,185 +1,200 @@
...
+
+  private void mockResponse(final String responseAsString, final Integer... statusCodes) {
+    final List<CloseableHttpResponse> responses =
+        Arrays.stream(statusCodes)
+            .map(
+                statusCode ->
+                    HttpClientTestFactory.createHttpResponse(responseAsString, statusCode))
+            .toList();
+
+    final AtomicInteger index = new AtomicInteger(0);
```
kuntzed, line 175: Is this just to have a final int or do we really need that this is atomic?

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/DefaultOAuth2TokenKeyServiceTest.java
```diff
@@ -1,185 +1,200 @@
...
+        .hasMessageContaining("Response Headers [testHeader: testValue]")
+        .hasMessageContaining("Http status code 400");
+  }
+
+  @Test
+  public void retrieveTokenKeys_errorOccursDuringRetry_throwsServiceException() throws IOException {
+    mockResponse(ERROR_MESSAGE, 500, 400);
+    final DefaultTokenClientConfiguration config = DefaultTokenClientConfiguration.getConfig();
+    config.setRetryEnabled(true);
+    config.setMaxRetryAttempts(1);
```
kuntzed, line 115: if this test is about the error then we do not need to reduce the number of attempts to 1, do we?

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/SpringOAuth2TokenKeyServiceTest.java
kuntzed, line 1: what about test checking the number of retries and that non-retryable codes do not trigger a retry?

kuntzed (changes requested):
looks good, i just have some minor remarks

token-client/src/main/java/com/sap/cloud/security/client/DefaultTokenClientConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Override
+  public Set<Integer> getRetryStatusCodes() {
+    return retryStatusCodes;
+  }
+
+  @Override
+  public void setRetryStatusCodes(final String retryStatusCodes) {
+    this.retryStatusCodes = parseRetryStatusCodes(retryStatusCodes);
+  }
```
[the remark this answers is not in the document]
NiklasHerrmann21: Moved the setter as well as the parser to the interface as a default method.

token-client/src/main/java/com/sap/cloud/security/client/TokenClientConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+   */
+  default void setRetryStatusCodes(final String retryStatusCodes) {
+    setRetryStatusCodes(parseRetryStatusCodes(retryStatusCodes));
+  }
+
+  private Set<Integer> parseRetryStatusCodes(final String retryStatusCodes) {
+    return Arrays.stream(Optional.ofNullable(retryStatusCodes).orElse("").split(","))
+        .map(String::trim)
+        .filter(s -> !s.isBlank())
+        .map(Integer::parseInt)
```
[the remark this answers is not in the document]
NiklasHerrmann21: Here we could discuss if we want it to simply crash at this point or if we want to catch that and rethrow the exception with the possibility of logging
@kuntzed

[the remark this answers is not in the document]
kuntzed: Possibly a good idea to log an error and maybe rethrow as another exception, e.g. IllegalStateException?

token-client/src/test/java/com/sap/cloud/security/client/AbstractTokenClientConfigurationTest.java
```diff
@@ -0,0 +1,67 @@
+package com.sap.cloud.security.client;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.jupiter.api.Test;
+
+public abstract class AbstractTokenClientConfigurationTest {
```
kuntzed, line 8: Do we really need to test the string parsing with both config implementations?

token-client/src/test/java/com/sap/cloud/security/xsuaa/client/SpringOAuth2TokenKeyServiceTest.java
```diff
@@ -59,11 +64,14 @@ public SpringOAuth2TokenKeyServiceTest() throws IOException {
   public void setUp() {
     restOperationsMock = Mockito.mock(RestOperations.class);
     cut = new SpringOAuth2TokenKeyService(restOperationsMock);
+    final SpringTokenClientConfiguration config = SpringTokenClientConfiguration.getConfig();
+    config.setRetryEnabled(false);
+    SpringTokenClientConfiguration.setConfig(config);
```
kuntzed, line 69: Why is this only required in the Spring tests?

kuntzed (changes requested):

token-client/src/test/java/com/sap/cloud/security/client/AbstractTokenClientConfigurationTest.java
NiklasHerrmann21: I think we should keep the string parsing tests in the abstract test class. This approach ensures that any changes to the default method or overwriting it in the implementations are detected early.
This way we might test redundantly but for me it makes the codebase easier to maintain in the long run.

kuntzed (approved):

PR #1776, opened 2025-04-28
added documentation of retry feature to README

NiklasHerrmann21:


token-client/README.md
```diff
@@ -325,6 +325,44 @@ OAuth2TokenResponse tokenResponse = tokenFlows.passwordTokenFlow()
                                     .execute();
 ```
 
+## Retry mechanism
+
+The retry feature (supported since version 3.6.0) uses two Configuration classes to handle its properties and overall activation.
+
+`DefaultTokenClientConfiguration` is used for Java EE applications and `SpringTokenClientConfiguration` is used for Spring Boot applications.
+
+Both disabled by default for all TokenFlow classes and TokenService classes and use a singleton pattern to ensure that the same instance is used throughout the application.
```
kuntzed, line 334: Both are disabled

kuntzed (changes requested):

kuntzed (approved):

kuntzed (approved):

PR #1848, opened 2025-09-09
Scicas 3580 ias token exchange

NiklasHerrmann21:


java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
```diff
@@ -0,0 +1,191 @@
...
+      final Mode mode,
+      final CloseableHttpClient httpClient,
+      final OAuth2ServiceConfiguration xsuaaConfig) {
+    this.iasConfig = Objects.requireNonNull(iasConfig, "iasConfig must not be null");
+    this.mode = Objects.requireNonNull(mode, "mode must not be null");
+    this.tokenService = new DefaultOAuth2TokenService(httpClient);
+    this.xsuaaConfig = xsuaaConfig;
+  }
+
+  /** Creates a new IAS to XSUAA extension. */
```
kuntzed: Sound like a constructor comment.

java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.http.impl.client.CloseableHttpClient;
+
+/**
+ * Exchanges IAS weak -> IAS strong, and optionally IAS strong -> XSUAA. Stores results as context
+ * attachments and (optionally) switches primary token.
+ */
+public class IasToXsuaaExtension implements ContextExtension {
+
+  public enum Mode {
+    NONE,
```
kuntzed: Who needs the NONE mode?

java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
```diff
@@ -0,0 +1,191 @@
...
+    TO_IAS, // weak IAS -> strong IAS
```
kuntzed: So in TO_IAS Mode the IasToXsuaaExtension is not doing the conversion to XSUAA? Maybe we need another name or have separate extensions for xsuaa and string ias? Is there a use case for only doing the first token exchange?

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -10,32 +10,57 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class IasTokenAuthenticator extends AbstractTokenAuthenticator {
+  private static final Logger LOG = LoggerFactory.getLogger(IasTokenAuthenticator.class);
+  private ContextExtensionsRegistry contextExtensionsRegistry;
 
-	@Override
-	public Token extractFromHeader(String authorizationHeader) {
+  public IasTokenAuthenticator withExtensionsRegistry(final ContextExtensionsRegistry registry) {
```
kuntzed: Why not simply always initializing with an empty registry?

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -10,32 +10,57 @@
...
+    final TokenAuthenticationResult validationResult = super.validateRequest(request, response);
+    if (validationResult.isAuthenticated()) {
+      if (contextExtensionsRegistry != null) {
+        try {
+          contextExtensionsRegistry.applyAll();
+        } catch (final Exception e) {
+          LOG.warn("Security context extension failed", e);
+        }
+        return TokenAuthenticatorResult.createAuthenticated(
+            Collections.emptyList(), SecurityContext.getToken());
```
kuntzed: Here, the original validation result is thrown away, assuming that it is anyway created in the same way as the new one. If that changes in the future, then this part would probably need to be adapted as well while being somewhat hidden.

Apart from that, this is the point that needs to be discussed since here we now return an authentication result for a token that was not validated at all, if the new extension is run.

Furthermore, in case the result is an XSUAA token, the new result diverges from that of the XsuaaTokenAuthenticator which includes the scopes of the xsuaa token instead of the empty list.

kuntzed (changes requested):
I wonder if we should really embed this token exchange to xsuaa into the IasTokenAuthenticator. Maybe we need a separate Authenticator class. If I understand the use case correctly, the applications that will use this are XSUAA applications. That is they will probably migrate away from the XsuaaTokenAuthencticator. So why should they migrate to the IasTokenAuthenticator if they afterwards expect to continute to work with XSUAA tokens.

Furthermore, I wonder if the use cases also cover the hybrid cases where applications may still receive XSUAA tokens. That is, maybe we need a HybridTokenAuthenticator that is able to handle IAS and XSUAA tokens.

java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
NiklasHerrmann21: Thats a good question. 
I wanted to leave both options open in case we might need a switch from weak -> strong without the XSUAA swap in the future so we wouldnt need a new implementation

java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
NiklasHerrmann21: That was still in from my first draft but since you need to explicitly declair that you want to exchange a token by extending the context, there would be no need for a "NONE" mode.
Therefore good to remove

java-security/src/main/java/com/sap/cloud/security/token/extension/IasToXsuaaExtension.java
kuntzed: ok, but with NONE gone and no use case for TO_IAS, we can drop the mode completely.

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
NiklasHerrmann21: good point, that would simplify the code further.
I will resolve this with the next commit.

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,206 @@
...
+      final OAuth2ServiceConfiguration xsuaaConfig) {
+    this.iasConfig = Objects.requireNonNull(iasConfig, "iasConfig must not be null");
+    this.tokenService = new DefaultOAuth2TokenService(httpClient);
+    this.xsuaaConfig = xsuaaConfig;
+  }
+
+  @Override
+  public TokenAuthenticationResult validateRequest(
+      final ServletRequest request, final ServletResponse response) {
+    final TokenAuthenticationResult iasAthenticationResult =
```
kuntzed, line 53: What if the incoming token is already an XSUAA token? If I remember correctly, then such cases need to be covered as well.

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,206 @@
...
+        iasTokenAuthenticator.validateRequest(request, response);
+    if (iasAthenticationResult.isAuthenticated()) {
+      try {
+        final Token incoming = SecurityContext.getToken();
+        final boolean isWeak = isWeakToken(incoming);
+        Token strongIas = incoming;
+        if (isWeak) {
+          LOG.info("Detected weak IAS token. Exchanging to strong IAS …");
+          strongIas = exchangeToStrongIas(incoming);
+          SecurityContext.setToken(strongIas);
```
kuntzed, line 63: Why do we write the ID Token to the context if it will be anyway replaced with the XSUAA token later?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,206 @@
...
+  private final OAuth2TokenService tokenService;
+
+  IasTokenAuthenticator iasTokenAuthenticator = new IasTokenAuthenticator();
+  XsuaaTokenAuthenticator xsuaaTokenAuthenticator = new XsuaaTokenAuthenticator();
+
+  public HybridTokenAuthenticator(
+      final OAuth2ServiceConfiguration iasConfig,
+      final CloseableHttpClient httpClient,
+      final OAuth2ServiceConfiguration xsuaaConfig) {
+    this.iasConfig = Objects.requireNonNull(iasConfig, "iasConfig must not be null");
```
kuntzed: Why is it ok for the xsuaaConfig to be null?

kuntzed (changes requested):

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -13,6 +13,7 @@
 import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import org.jetbrains.annotations.ApiStatus;
```
kuntzed: Do we really need to force all our consumers into this dependency?

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -78,6 +78,21 @@ public static void setToken(Token token) {
...
+   *
+   * @param token token to be saved.
+   */
+  @ApiStatus.Internal
+  public static void overwriteToken(final Token token) {
+    LOGGER.debug(
+        "Sets token of service {} to SecurityContext (thread-locally).",
+        token != null ? token.getService() : "null");
+    tokenStorage.set(token);
+  initialTokenStorage.set(token);
```
kuntzed: So this does the same as setToken()? Shouldn't this leave the initial token untouched?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,129 @@
...
+    this.xsuaaConfig = xsuaaConfig;
+  }
+
+  @Override
+  public TokenAuthenticationResult validateRequest(
+      final ServletRequest request, final ServletResponse response) {
+    final TokenAuthenticationResult authenticationResult =
+        iasTokenAuthenticator.validateRequest(request, response);
+    if (authenticationResult.isAuthenticated()) {
+      if (SecurityContext.getToken() instanceof XsuaaToken) {
```
kuntzed: How can the ias authenticator validation result in an xsuaa token?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,129 @@
...
+        return authenticationResult;
+      }
+      try {
+        SecurityContext.registerIdTokenExtension(
+            new DefaultIdTokenExtension(tokenService, iasConfig));
+        final Token idToken = SecurityContext.getIdToken();
+        if (idToken == null) {
```
kuntzed: maybe we should spend a comment to explain this situation. Does null mean that the id token exchange failed and we simply return the initial token?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+    return Token.create(
+        "Bearer "
+            + tokenService.retrieveAccessTokenViaJwtBearerTokenGrant(
+                tokenEndpoint, identity, idToken.getTokenValue(), params, false, zid));
+  }
+
+
+  @Override
+  protected OAuth2ServiceConfiguration getServiceConfiguration() {
```
kuntzed, line 109: I see no difference to the inherited method...

kuntzed (changes requested):
Let's discuss if we need something like a mode for the hybrid authenticator.

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
NiklasHerrmann21: i can also add a javadoc to avoid this dependency

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
NiklasHerrmann21: Yes this was a copy paste error.
Initial token and id token should stay

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
NiklasHerrmann21: We should talk about this case further
What happens if a XSUAA Token is in the context and gets validated?
Do we need a switch in here to call the right authenticator?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
NiklasHerrmann21: Yes, this was my thought.
if theres an OAuth2ServiceException during the retrieval of the id token null will be returned 
so in that case we cant switch to XSUAA and should return the initial result?
Or should we throw an error?

java-api/pom.xml
```diff
@@ -116,7 +116,12 @@
 			<scope>test</scope>
 			<version>${slf4j.api.version}</version>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>26.0.2</version>
+        </dependency>
```
kuntzed: Did you check, if this is really required?

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -78,6 +77,18 @@ public static void setToken(Token token) {
     idTokenStorage.remove();
   }
 
+  /**
+   * Saves the token thread wide. only used in special cases where the token for internal usage.
```
kuntzed: the sentence looks incomplete

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
```diff
@@ -196,21 +207,22 @@ public static List<String> getServicePlans() {
...
+    // the header format contains a comma-separated list of quoted plan names, e.g. "plan1","plan
+    // \"two\"","plan3"
+    final String[] planParts =
+        servicePlansHeader.trim().split("\\s*,\\s*"); // split by <whitespaces>,<whitespaces>
+
+    // remove " around plan names
+    final List<String> plans =
+        Arrays.stream(planParts)
+            .map(plan -> plan.substring(1, plan.length() - 1))
+            .collect(Collectors.toList());
```
kuntzed, line 226: Is this just reformatting from your ide?

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
```diff
@@ -14,38 +14,39 @@
...
+      throw new IllegalStateException("There must be a service configuration.");
+    }
+    return config;
+  }
+
+  @Nullable
+  @Override
+  protected OAuth2ServiceConfiguration getOtherServiceConfiguration() {
+    return null;
+  }
```
kuntzed, line 51: only format change again?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,138 @@
...
+          if (idToken == null) {
+            throw new IllegalStateException("Can not retrieve XSUAA Token. No ID Token present!");
+          }
+          final Token xsuaaToken = exchangeToXsuaa(idToken);
+          final TokenAuthenticationResult xsuaaAuthenticationResult =
+              xsuaaTokenAuthenticator.authenticated(xsuaaToken);
+          SecurityContext.overwriteToken(xsuaaToken);
+          return xsuaaAuthenticationResult;
+        } catch (final OAuth2ServiceException e) {
+          return authenticationResult;
```
kuntzed: throw exception?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+      } else {
+        return authenticationResult;
```
kuntzed: throw exception?

java-api/pom.xml
NiklasHerrmann21: Just removed it

java-api/src/main/java/com/sap/cloud/security/token/SecurityContext.java
NiklasHerrmann21: Jep, just reformatting that happened during the implementation

java-security/src/main/java/com/sap/cloud/security/servlet/IasTokenAuthenticator.java
NiklasHerrmann21: Jep, reformatting from when i first implemented the change within the ias authenticator before we moved to a hybrid authenticator

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
NiklasHerrmann21: As discussed via teams i will return an unauthenticated result in case of an error

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
NiklasHerrmann21: same as above

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -22,70 +22,94 @@
...
+    // If token is already an XSUAA token, delegate to XSUAA authenticator
+    if (isXsuaaToken(decodedJwt)) {
+      return xsuaaTokenAuthenticator.validateRequest(httpRequest, response);
+    }
+
+    // Otherwise, treat it as an IAS token
+    final TokenAuthenticationResult iasResult =
+        iasTokenAuthenticator.validateRequest(httpRequest, response);
+    if (!iasResult.isAuthenticated()) {
+      return unauthenticated(
```
kuntzed: maybe it's good enough to simply return the iasResult?

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -22,70 +22,94 @@
...
+    try {
+      final Token idToken = SecurityContext.getIdToken();
+      if (idToken == null) {
+        return unauthenticated("Missing IAS ID Token. Cannot exchange for XSUAA Token.");
       }
+      final Token xsuaaToken = exchangeToXsuaa(idToken);
+      SecurityContext.overwriteToken(xsuaaToken);
+      return xsuaaTokenAuthenticator.authenticated(xsuaaToken);
+    } catch (OAuth2ServiceException e) {
+      return unauthenticated("Unexpected error occurred: " + e.getMessage());
```
kuntzed: "Unexpected error during exchange from ID token to XSUAA token: "
Out future selfs will probably be grateful for this context information. ;-)

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -22,70 +22,94 @@
...
+    } catch (IllegalStateException e) {
+      return unauthenticated("Token exchange failed: " + e.getMessage());
```
kuntzed: do we need two different catch blocks here?

kuntzed (changes requested):

java-security/src/main/java/com/sap/cloud/security/servlet/HybridTokenAuthenticator.java
```diff
@@ -105,10 +103,9 @@ public TokenAuthenticationResult validateRequest(
       final Token xsuaaToken = exchangeToXsuaa(idToken);
       SecurityContext.overwriteToken(xsuaaToken);
       return xsuaaTokenAuthenticator.authenticated(xsuaaToken);
-    } catch (OAuth2ServiceException e) {
-      return unauthenticated("Unexpected error occurred: " + e.getMessage());
-    } catch (IllegalStateException e) {
-      return unauthenticated("Token exchange failed: " + e.getMessage());
+    } catch (Exception e) {
```
kuntzed: catch (OAuth2ServiceException | IllegalStateException | e) ?

PR #1936, opened 2026-04-01
feat: Add token-client-spring-3 module for Spring Boot 3.x compatibility

kuntzed:
## Summary

- Add new `token-client-spring-3` module compiled against Spring Framework 6.x for Spring Boot 3.x compatibility
- Centralize Spring Boot 3.x versions in parent POM using `legacy3.spring.*` properties
- Update migration guide with correct Spring Boot 3.x starter module name

## Background

The `token-client-spring` module is compiled against Spring Framework 7.x (Spring Boot 4.x), which introduces a new `HttpEntity` constructor signature. When used with Spring Boot 3.x applications, this causes `NoSuchMethodError` at runtime because Spring 6.x doesn't have the same constructor.

This PR adds a separate module (`token-client-spring-3`) that is compiled against Spring 6.x, ensuring bytecode compatibility with Spring Boot 3.x applications.

## Changes

| File | Change |
|------|--------|
| `pom.xml` | Add `legacy3.spring.*` properties and new module |
| `token-client-spring-3/` | New module (same source as token-client-spring, compiled against Spring 6.x) |
| `spring-security-3/pom.xml` | Use centralized `legacy3.*` properties |
| `spring-security-3-starter/pom.xml` | Use centralized `legacy3.*` properties |
| `bom/pom.xml` | Add new artifact |
| `MIGRATION_4.0.md` | Update module name references |

## Test plan

- [x] `mvn clean install -pl token-client-spring-3` builds successfully
- [x] All tests pass in token-client-spring-3 module
- [x] Verified bytecode uses Spring 6.x compatible `HttpEntity` constructor
- [x] Full build passes

### DIFF
--- a/spring-security-3/pom.xml
+++ b/spring-security-3/pom.xml
@@ -50,12 +50,12 @@
 
 	<properties>
 		<!-- Override parent's Spring Boot 4.x with Spring Boot 3.x for legacy compatibility -->
-		<spring.boot.version>3.5.9</spring.boot.version>
-		<spring.core.version>6.2.15</spring.core.version>
-		<spring.security.version>6.5.7</spring.security.version>
-		<spring.security.oauth2.version>6.5.7</spring.security.oauth2.version>
-		<reactor.version>3.8.2</reactor.version>
-		<reactor.test.version>3.8.2</reactor.test.version>
+		<spring.boot.version>${legacy3.spring.boot.version}</spring.boot.version>
+		<spring.core.version>${legacy3.spring.core.version}</spring.core.version>
+		<spring.security.version>${legacy3.spring.security.version}</spring.security.version>
+		<spring.security.oauth2.version>${legacy3.spring.security.oauth2.version}</spring.security.oauth2.version>
+		<reactor.version>${legacy3.reactor.version}</reactor.version>
+		<reactor.test.version>${legacy3.reactor.version}</reactor.test.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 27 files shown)


spring-security-3/pom.xml
```diff
@@ -156,7 +156,7 @@
 		<dependency>
 			<groupId>io.projectreactor</groupId>
 			<artifactId>reactor-core</artifactId>
-			<version>3.8.2</version>
+			<version>${reactor.version}</version>
```
NiklasHerrmann21, line 159: Why not use the legacy3.reactor.version?

spring-security-3/pom.xml
```diff
@@ -182,19 +182,19 @@
...
 			<scope>test</scope>
-			<version>3.5.9</version>
+			<version>${spring.boot.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>io.projectreactor</groupId>
 			<artifactId>reactor-test</artifactId>
 			<scope>test</scope>
-			<version>3.8.2</version>
+			<version>${reactor.test.version}</version>
```
NiklasHerrmann21, line 197: Same thing here, i think we need legacy3.spring.boot.version and legacy3.reactor.version as we otherwise would face issues as the "higher" versions would be used

NiklasHerrmann21 (approved):

spring-security-3/pom.xml
kuntzed: overridden via properties

PR #1939, opened 2026-04-02
Bugfix/4.0.2

NiklasHerrmann21:


finkmanAtSap (approved):
Token is already validated at this point, so it is safe to call the issuer URL 👍🏻

PR #1971, opened 2026-06-09
feat: IAS token flow with tenant host resolution

NiklasHerrmann21:
## Summary

Adds a multi-tenant-capable IAS token flow that resolves the correct subscriber token endpoint at runtime via the BTP tenant API, instead of hard-coding the provider's IAS host. Resolved subdomains are cached with a configurable Caffeine cache.

- New `IasTenantHostResolver` queries `/sap/rest/tenantLoginInfo?id={tenantId}` on the BTP tenant API and extracts the subscriber's IAS subdomain from the returned `token_endpoint`.
- `IasTokenFlows` integrates the resolver: when an `app_tid` is supplied, the configured IAS host is rewritten to the subscriber's host before token requests; otherwise the provider host is used unchanged.
- Auto-detection of the BTP tenant API base URI from the IAS service binding (no manual wiring needed when the binding exposes it).
- Cache is now governed by `IasTenantHostCacheConfiguration` (enabled / TTL / max size). Default: enabled, 1h TTL, 1000 entries. Spring Boot consumers can bind via `sap.spring.security.ias.tenant-host-cache.*`.

## Commits

- `feat: Add IAS token flow support with tenant host resolution`
- `refactor: Auto-detect btp-tenant-api from IAS service binding`
- `feat: Make IAS tenant host cache configurable`

## Test plan

- [ ] `mvn -pl token-client test` is green (17 IAS-tenant-host tests pass locally)
- [ ] `mvn -pl spring-security compile` succeeds
- [ ] Verify against an IAS binding *with* `btp-tenant-api` set: subscriber tenant token request hits subscriber subdomain
- [ ] Verify against an IAS binding *without* `btp-tenant-api` set: falls back to provider host (no resolver constructed)
- [ ] Verify Spring Boot binding: setting `sap.spring.security.ias.tenant-host-cache.enabled=false` skips caching and every resolve hits the BTP API
- [ ] Verify cache eviction by TTL with a short `ttl` value

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -3,9 +3,6 @@
 ---
 name: Maven Build main
 
-env:
-  NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
-
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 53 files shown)


NiklasHerrmann21:
Blocked - waiting for test environment to test the tenant host retrieval

PR #1983, opened 2026-06-22
feat: Expose sap_id_type claim via SapIdToken#getIdType

NiklasHerrmann21:
## Summary
- New `SapIdToken#getIdType()` returning a typed `SapIdType` enum (`USER` / `APP`) backed by the `sap_id_type` JWT claim issued by SAP Cloud Identity Service.
- New `TokenClaims.SAP_ID_TYPE` constant.
- `DefaultIdTokenExtension#isTechnicalUser` now prefers the claim and falls back to the `sub == azp` heuristic for tokens issued before the claim was introduced — no behavior change for pre-`sap_id_type` bindings, but new tokens drive the right decision when the heuristic would have been wrong.

## Why
The Node library exposes the same claim (`token.idType`) and uses it to replace the `sub === azp` heuristic at decision points like `DefaultIdTokenExtension._getIdToken` and `XsuaaLegacyExtension.appliesTo`. This brings the Java library in line.

## Design notes
- `SapIdType` lives in `java-api` next to `TokenClaims` so it can be referenced from both `java-api` and `java-security`.
- The enum returns `null` for both an absent claim and an unknown value (e.g. a future third principal type). This keeps consumers forward-compatible without throwing on tokens issued by newer IAS versions.
- `isTechnicalUser` keeps the legacy `sub == azp` path as a fallback to avoid breaking technical-user detection for tokens that don't yet carry `sap_id_type`.

## Test plan
- [x] `mvn -pl java-api,java-security -am test` — 330 tests green, including new coverage:
  - `SapIdTokenTest`: `getIdType` for `user`, `app`, missing claim, unknown value
  - `DefaultIdTokenExtensionTest`: `sap_id_type=app` is treated as technical user; `sap_id_type=user` overrides the `sub == azp` heuristic
- [x] `mvn install -DskipTests` (full repo compile + javadoc) — green

## Release
CHANGELOG entry added under `4.1.0` (minor bump for additive feature). The pom version bump will follow in a separate `chore: Release 4.1.0` commit once any in-flight Renovate PRs are merged.

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
@@ -27,7 +27,8 @@ import org.slf4j.LoggerFactory;
  * <ul>
  *   <li>If cached ID Token is still valid, it is returned as is
  *   <li>If the current token is already an ID token, it is returned as-is.
- *   <li>If the token belongs to a technical user (where {@code sub == azp}), an exception is
+ *   <li>If the token belongs to a technical user (claim {@code sap_id_type} = {@code "app"}, or
+ *       {@code sub == azp} for pre-{@code sap_id_type} tokens), an exception is
  *       thrown.
  *   <li>If the token is an access token, it will be exchanged for an ID token using the configured
  *       IAS service credentials.
@@ -115,14 +116,21 @@ public class DefaultIdTokenExtension implements IdTokenExtension {
   /**
    * Determines whether the token represents a technical user.
    *
-   * <p>A token is considered to belong to a technical user if the {@code sub} (subject) claim
-   * equals the {@code azp} (authorized party / client ID) claim.
+   * <p>Prefers the {@code sap_id_type} claim ({@link SapIdType#APP}) when present. For tokens
+   * issued before the claim was introduced, falls back to comparing {@code sub} with
+   * {@code azp}.
    *
    * @param token the token to inspect
    * @return {@code true} if the token belongs to a technical user
    */
   private boolean isTechnicalUser(Token token) {
-    String subject = token.getClaimAsString("sub");
+    if (token instanceof SapIdToken idToken) {
+      SapIdType idType = idToken.getIdType();
+      if (idType != null) {
+        return idType == SapIdType.APP;
+      }
+    }
+    String subject = token.getClaimAsString(TokenClaims.SUBJECT);
     String azp = token.getClientId();
     if (subject == null || azp == null || subject.isBlank() || azp.isBlank()) {
       return false;
--- a/java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
@@ -27,7 +27,7 @@ import org.slf4j.LoggerFactory;
  * <ul>
  *   <li>If cached ID Token is still valid, it is returned as is
  *   <li>If the current token is already an ID token, it is returned as-is.
- *   <li>If the token belongs to a technical user (claim {@code sap_id_type} = {@code "app"}, or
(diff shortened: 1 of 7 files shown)


java-security/src/main/java/com/sap/cloud/security/token/SapIdToken.java
```diff
@@ -59,4 +59,17 @@ public String getIssuer() {
...
+	 * Identity Service to disambiguate human end-users ({@link SapIdType#USER}) from
+	 * technical/application principals ({@link SapIdType#APP}).
+	 *
+	 * @return the resolved {@link SapIdType}, or {@code null} if the claim is absent or carries
+	 * 		an unknown value (e.g. a future principal type).
+	 */
+	@Nullable
+	public SapIdType getIdType() {
+		return SapIdType.fromClaimValue(getClaimAsString(TokenClaims.SAP_ID_TYPE));
+	}
```
finkmanAtSap: The enum return type is the cleanest approach but it has the problem that it is not forward-compatible if new claim values are added, e.g. `"agent"` perhaps. You will need to do new releases and consumers will need to update before being able to read the claim correctly. With the old version, it will look like the claim is missing which might be misleading.

I would go for a `String` return value and export String constants for the currently known values, i.e. `APP` and `USER` that consumers can compare with today. It's less nice than comparing with an enum but it's honest about the fact that the claim could have more than these two values and allows programming already against that.

This is important when high-lever modules in the library stack like CAP use this feature: they cannot simply set a minimum version of the BTP sec libs for applications and begin to use newer enum values. Instead, they need to work also when older versions of the library are installed and where no enum value exists for new claim values.

java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
```diff
@@ -115,14 +116,21 @@ private boolean isAccessToken(Token token) {
...
-   * <p>A token is considered to belong to a technical user if the {@code sub} (subject) claim
-   * equals the {@code azp} (authorized party / client ID) claim.
+   * <p>Prefers the {@code sap_id_type} claim ({@link SapIdType#APP}) when present. For tokens
+   * issued before the claim was introduced, falls back to comparing {@code sub} with
+   * {@code azp}.
    *
    * @param token the token to inspect
    * @return {@code true} if the token belongs to a technical user
    */
   private boolean isTechnicalUser(Token token) {
```
finkmanAtSap, line 131: Not sure how it is implemented in Node.js (probably like this) but I think it would be best to be conservative and demand `"user"` before attempting a token exchange. Theoretically, in this case, there could be new types introduced later for which the extension should also do the exchange but it does not do it because it requires `"user"`. However, this is better than the alternative where the extension checks for `"app"` and attempts an exchange in all other cases, even when it should not (e.g. when new non-named-user values like `"agent"` are introduced).

finkmanAtSap (changes requested):
I have some doubts about the strongly types enum approach. Let's discuss whether you see it the same way as me or not.

java-security/src/main/java/com/sap/cloud/security/token/SapIdToken.java
NiklasHerrmann21: Good point — updated in 5f3e307. `getIdType()` now returns the raw claim string, and `SapIdType` is a holder for the string constants `USER` / `APP`. Consumers can compare against those today and still see any future IAS value verbatim without a library upgrade.

java-security/src/main/java/com/sap/cloud/security/token/DefaultIdTokenExtension.java
NiklasHerrmann21: Agreed — flipped to the conservative check in 5f3e307. `isTechnicalUser` now returns `true` for any `sap_id_type` value that is not exactly `"user"`, so future values like `"agent"` are refused rather than silently exchanged. Added a test (`resolveToken_sapIdTypeUnknownClaim_conservativelyTreatedAsTechnicalUser`) to lock the behavior in.

PR #1991, opened 2026-06-25
feat: Support RS384/RS512, PS256/PS384/PS512, ES256/ES384/ES512 JWT signature algorithms

NiklasHerrmann21:
## Summary
Adds support for the full RFC 7518 §3.1 set of JWT signature algorithms in `JwtSignatureValidator`, beyond the pre-existing `RS256`:

- `RS384`, `RS512` — RSASSA-PKCS1-v1_5 with SHA-384 / SHA-512 (RFC 7518 §3.3)
- `PS256`, `PS384`, `PS512` — RSASSA-PSS with SHA-256 / SHA-384 / SHA-512 (RFC 7518 §3.5)
- `ES256`, `ES384`, `ES512` — ECDSA with P-256 / P-384 / P-521 (RFC 7518 §3.4)

`RS256` continues to work unchanged.

## What changed
- `JwtSignatureAlgorithm` (enum) now declares all nine algorithms with the JCA name passed to `Signature.getInstance(...)`, the `PSSParameterSpec` for PSS entries, and the curve name + coordinate length for EC entries. For ECDSA the JCA name `SHA*withECDSAinP1363Format` is used so `Signature#verify` accepts the raw `R||S` concatenation mandated by RFC 7518 §3.4 directly, without DER transcoding.
- `JwtSignatureValidator#validateSignature` calls `Signature#setParameter(...)` when the selected algorithm carries a parameter spec. RSA-PKCS1 and ECDSA paths need no parameter setup.
- **JWK key material is now carried by a sealed `KeyMaterial` type** (records `Rsa`, `Ec`, `Pem`) rather than eight nullable constructor parameters on `JsonWebKeyImpl`. Each record owns its key-construction logic; `JsonWebKeyImpl#getPublicKey` just delegates. Adding a further JWK key type (e.g. OKP for EdDSA) means adding one record, not another optional constructor parameter.
- EC public keys are built from JWK `crv`/`x`/`y` with strict validation: the curve must match the algorithm and each coordinate must have the exact octet length per RFC 7518 §6.2.1. A short or truncated coordinate would otherwise produce a key on the wrong subgroup and silently succeed.
- JWK curve names `P-256`/`P-384`/`P-521` are mapped to the NIST names `secp256r1`/`secp384r1`/`secp521r1` before being passed to `AlgorithmParameters.init(ECGenParameterSpec)`. SunEC accepts only the NIST names for the larger two curves; without this mapping ES384/ES512 would fail at runtime on the reference JCA provider.
- `JwtSignatureAlgorithm#fromType` iterates enum declaration order and returns the first `kty` match — `RS256` for `RSA`, `ES256` for `EC`. The choice is only advisory; downstream key construction cross-checks the concrete JWK parameters against the returned algorithm, so a mismatched default surfaces as a clear error rather than a silently wrong key.

## What did **not** change
- The validator still only accepts algorithms it has been told about; an unknown `alg` in the JWT header is rejected with the existing `JWT token validation with signature algorithm '...' is not supported.` error.
- The PEM-encoded-key path used by the XSUAA JKU fallback continues to work — it is now expressed as `KeyMaterial.Pem` rather than a dedicated constructor branch.

## Test plan
- [x] `mvn -pl java-security test` — green (371 tests, 0 failures, 2 pre-existing `@Disabled` skipped)
- [x] `mvn install -DskipTests` (full repo compile) — green
- [x] `JwtSignatureAlgorithmTest` (parametrized over all 9 algorithms) — happy path + tamper rejection in isolation
- [x] `SapIdJwtSignatureValidatorAlgorithmsTest` — end-to-end through the real `SapIdJwtSignatureValidator` with mocked transport: generates a fresh key pair per algorithm, builds a synthetic JWKS, signs a token, validates it. 9 × happy + 9 × tamper.
- [x] `JsonWebKeySetFactoryTest` extended with PS256, ES256/384/512 parsing, plus EC curve-mismatch and coordinate-length-mismatch error cases
- [x] Manually validated real IAS-issued ES256 and PS256 tokens (with their matching JWKS documents from the IAS tenant) end-to-end through `JwtSignatureValidator#validateSignature` — proves the JWK parser + JCA glue accept actual IAS signature material, not just synthetic keys
- [x] All keys/tokens in the committed test suite are generated at test time — no real tenant material in the repo
- [x] Algorithm-confusion safety: the algorithm is taken from the JWT header but only used as a lookup key into the enum; we never trust an attacker-supplied algorithm name verbatim, and the PSS parameter spec / EC curve are fixed per enum entry rather than driven by header values

## Companion ticket
Lifts the supported-algorithm set to full RFC 7518 §3.1 parity with the node.js library.

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureAlgorithm.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureAlgorithm.java
@@ -83,12 +83,20 @@ public enum JwtSignatureAlgorithm {
 
 	/**
 	 * Returns the default algorithm for the given JWK key type. Used as a fallback when a JWK
-	 * entry omits its {@code alg}. Falls back to {@link #RS256} for {@code RSA} keys; returns
-	 * {@code null} for unknown types.
+	 * entry omits its {@code alg}.
+	 * <p>
+	 * When multiple algorithms share a {@code kty}, the first match in enum declaration order
+	 * wins — currently {@link #RS256} for {@code RSA}. Preserve that order when adding new
+	 * algorithms (e.g. EC): put the canonical RFC 7518 default at the top of its {@code kty}
+	 * group. The choice is only advisory; downstream key construction still cross-checks the
+	 * concrete JWK parameters (curve, coordinate length, …) against the returned algorithm,
+	 * so a mismatched default produces a clear error rather than a silently wrong key.
 	 */
 	public static JwtSignatureAlgorithm fromType(String type) {
-		if ("RSA".equals(type)) {
-			return RS256;
+		for (JwtSignatureAlgorithm algorithm : values()) {
+			if (algorithm.type.equals(type)) {
+				return algorithm;
+			}
 		}
 		return null;
 	}
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureAlgorithm.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureAlgorithm.java
@@ -16,13 +16,19 @@ import java.security.spec.PSSParameterSpec;
  * <p>
  * Each entry carries:
  * <ul>
- *   <li>the {@code kty} JWK key type ({@code RSA} for RS* and PS* families),
+ *   <li>the {@code kty} JWK key type ({@code RSA} for RS* and PS* families, {@code EC} for ES*),
  *   <li>the {@code alg} value as it appears in JWK and JWT headers,
  *   <li>the corresponding standard JCA signature algorithm name passed to
  *       {@link java.security.Signature#getInstance(String)},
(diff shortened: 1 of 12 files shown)


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JwtSignatureAlgorithm.java
```diff
@@ -43,12 +81,15 @@ public static JwtSignatureAlgorithm fromValue(String value) {
 		return null;
 	}
 
+	/**
+	 * Returns the default algorithm for the given JWK key type. Used as a fallback when a JWK
+	 * entry omits its {@code alg}. Falls back to {@link #RS256} for {@code RSA} keys; returns
```
kuntzed: I only see the fallbacks to RS256 and null. Don't know what happened before for type != "RSA", but does not seem to handle other types than RSA.

kuntzed (approved):

kuntzed (approved):

PR #1995, opened 2026-06-29
feat: Support ES256/ES384/ES512 JWT signature algorithms

NiklasHerrmann21:
## Summary

- Adds ECDSA support (ES256/ES384/ES512) to `JwtSignatureValidator`, completing the RFC 7518 §3.1 algorithm set on top of the RS*/PS* families added in #<rs-pr>. Builds on `feature/jwt-rsa-algorithms`.
- Uses the JCA name `SHA*withECDSAinP1363Format` so the raw `R||S` signature format mandated by RFC 7518 §3.4 is accepted directly, without DER transcoding.
- Constructs EC public keys from JWK `crv`/`x`/`y` with strict validation: the curve must match the algorithm and each coordinate must have the exact octet length per RFC 7518 §6.2.1.
- `JwtSignatureAlgorithm.fromType(\"EC\")` deliberately returns `null` — unlike RSA, EC keys are curve-bound and no safe default exists when a JWK omits its `alg` field.

## Implementation notes

- JWK curve names `P-256`/`P-384`/`P-521` are mapped to the NIST names `secp256r1`/`secp384r1`/`secp521r1` before being passed to `AlgorithmParameters.init(ECGenParameterSpec)`. SunEC accepts only the NIST names for the larger two curves; without this mapping ES384/ES512 would fail at runtime on the reference JCA provider.
- The PEM-encoded-key path in `JsonWebKeyImpl` is unchanged; EC support is implemented via a new `createECPublicKey` branch using `ECPublicKeySpec`.
- `JsonWebKeyImpl` constructor is overloaded — the existing 5-arg signature is preserved for backwards compatibility, EC fields go through a new 8-arg overload that only `JsonWebKeySetFactory` calls.

## Test plan

- [x] `JwtSignatureAlgorithmTest` (existing, now parametrized over all 9 algorithms) — happy path + tamper rejection in isolation.
- [x] New `SapIdJwtSignatureValidatorAlgorithmsTest` — end-to-end through the real `SapIdJwtSignatureValidator` with mocked transport: generates a fresh key pair per algorithm, builds a synthetic JWKS, signs a token, validates it. 9 × happy + 9 × tamper.
- [x] `JsonWebKeySetFactoryTest` extended with PS256, ES256/384/512 parsing, plus EC curve-mismatch and coordinate-length-mismatch error cases.
- [x] All keys/tokens are generated at test time — no real tenant material in the repo.
- [x] `mvn -pl java-security test` → 370 tests, 0 failures, 2 pre-existing `@Disabled` skipped.

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JsonWebKeyImpl.java
+++ b/java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JsonWebKeyImpl.java
@@ -8,55 +8,23 @@ package com.sap.cloud.security.token.validation.validators;
 import com.sap.cloud.security.xsuaa.Assertions;
 
 import jakarta.annotation.Nullable;
-import java.math.BigInteger;
-import java.security.AlgorithmParameters;
-import java.security.KeyFactory;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
-import java.security.spec.ECParameterSpec;
-import java.security.spec.ECPoint;
-import java.security.spec.ECPublicKeySpec;
-import java.security.spec.ECGenParameterSpec;
 import java.security.spec.InvalidKeySpecException;
-import java.security.spec.InvalidParameterSpecException;
-import java.security.spec.KeySpec;
-import java.security.spec.RSAPublicKeySpec;
-import java.security.spec.X509EncodedKeySpec;
-import java.util.Base64;
 import java.util.Objects;
-
-import static com.sap.cloud.security.token.validation.validators.JsonWebKeyConstants.BEGIN_PUBLIC_KEY;
-import static com.sap.cloud.security.token.validation.validators.JsonWebKeyConstants.END_PUBLIC_KEY;
 
 class JsonWebKeyImpl implements JsonWebKey {
 	private final JwtSignatureAlgorithm keyAlgorithm;
 	private final String keyId;
-	private final String pemEncodedPublicKey;
-	private final String modulus;
-	private final String publicExponent;
-	private final String curve;
-	private final String xCoordinate;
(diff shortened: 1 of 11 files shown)


java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JsonWebKeyImpl.java
```diff
@@ -28,16 +34,28 @@ class JsonWebKeyImpl implements JsonWebKey {
 	private final String pemEncodedPublicKey;
 	private final String modulus;
 	private final String publicExponent;
+	private final String curve;
```
kuntzed: so the JsonWebKeyImpl becomes a master object with all parameters that any algorithm needs?
maybe, we shoud at least use naming conventions to separate which parameter is for which algorithm? or we introduce some sort of parameter encapsulation?

kuntzed (changes requested):

java-security/src/main/java/com/sap/cloud/security/token/validation/validators/JsonWebKeyImpl.java
NiklasHerrmann21: Thats a good point. I reworked the JsonWebKeyImpl to work with an sealed interface as a parameter to reduce the number of params / complexity.

NiklasHerrmann21:
Superseded by PR #1991 — the ES256/ES384/ES512 support commit and the KeyMaterial refactor have been cherry-picked onto `feature/jwt-rsa-algorithms` so that all algorithm work lands in a single PR against `main`. Closing without merge; `feature/jwt-ec-algorithms` will be deleted separately.

PR #1996, opened 2026-07-02
fix: Sanitize x-client_cert header for IAS proof-token validation (4.0.8)

NiklasHerrmann21:
## Summary

Fixes a regression introduced in 4.0.0 that breaks IAS proof-token validation in Kyma/Istio environments with `credential-type: X509_GENERATED`.

- `SapIdJwtSignatureValidator` was setting the `x-client_cert` request header directly from `X509Certificate#getPEM()`. When the certificate originates from Istio's `x-forwarded-client-cert` (XFCC) header, the PEM includes `-----BEGIN CERTIFICATE-----` / `-----END CERTIFICATE-----` delimiters and CR/LF line breaks.
- Since 4.0.0, the token-client HTTP transport uses Java 11's `HttpClient` (`JavaHttpClientAdapter`), which enforces RFC 7230 and rejects any header value containing CR/LF with `IllegalArgumentException` (surfaced as `"Token signature can not be validated because: invalid header value: <PEM>"`).
- Result: every authenticated request behind Istio failed proof-token validation.

## Fix

- New `X509Certificate#getPEMHeaderValue()` accessor that returns bare base64-encoded DER (PEM delimiters and all whitespace stripped). The IAS JWKS endpoint accepts this form. `getPEM()`'s existing contract is unchanged so other callers aren't affected.
- `SapIdJwtSignatureValidator` now uses `getPEMHeaderValue()` when populating the `x-client_cert` header.
- Defensive check in `JavaHttpClientAdapter`: any header value containing CR or LF fails fast with a clear message naming the offending header key (not the value, which may be sensitive), so future regressions surface as an actionable transport-layer error instead of a misleading `"invalid header value: <full PEM dumped>"`.

## Release

Prepares 4.0.8:
- Version bumped 4.0.7 → 4.0.8 across all modules, READMEs and samples.
- CHANGELOG updated. Dependency updates and the `junit-bom` import fix that were previously staged for 4.1.0 are now part of 4.0.8.

## Test plan

- [x] `mvn -pl java-security,token-client test` — 327 tests, 0 failures, 0 errors, 2 skipped (pre-existing), BUILD SUCCESS
- [x] `X509CertificateTest`: new tests for multi-line PEM and XFCC-wrapped URL-encoded PEM inputs
- [x] `SapIdJwtSignatureValidatorTest`: new test feeding a multi-line PEM cert through the proof-token path, asserting the captured `x-client_cert` header contains no `-----BEGIN`, whitespace, `\n`, or `\r`
- [x] `JavaHttpClientAdapterTest`: new file with tests for `\n` and `\r` rejection with clear error message
- [ ] Manual verification behind Istio in Kyma with `credential-type: X509_GENERATED`

### DIFF
--- a/java-security/src/main/java/com/sap/cloud/security/x509/X509Certificate.java
+++ b/java-security/src/main/java/com/sap/cloud/security/x509/X509Certificate.java
@@ -142,11 +142,22 @@ public class X509Certificate implements Certificate {
 	}
 
 	/**
-	 * @return the certificate as bare base64-encoded DER, with PEM delimiters and all whitespace
-	 * 		removed. Safe to use as an HTTP header value (RFC 7230 forbids CR/LF in header values).
+	 * @return the leaf certificate as bare base64-encoded DER, safe to use as an HTTP header
+	 * 		value (RFC 7230 forbids CR/LF in header values). If the PEM contains multiple
+	 * 		concatenated certificates only the first is kept — consistent with
+	 * 		{@link #getSubjectDN()} and {@link #getThumbprint()}, which also operate on the first
+	 * 		(leaf) certificate.
 	 */
-	public String getPEMHeaderValue() {
-		return this.pem
+	public String getLeafCertificateAsHeaderValue() {
+		String pem = this.pem;
+		int begin = pem.indexOf("-----BEGIN CERTIFICATE-----");
+		if (begin >= 0) {
+			int end = pem.indexOf("-----END CERTIFICATE-----", begin);
+			if (end > begin) {
+				pem = pem.substring(begin, end + "-----END CERTIFICATE-----".length());
+			}
+		}
+		return pem
 				.replace("-----BEGIN CERTIFICATE-----", "")
 				.replace("-----END CERTIFICATE-----", "")
 				.replaceAll("\\s", "");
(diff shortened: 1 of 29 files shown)


kuntzed (dismissed):

java-security/src/main/java/com/sap/cloud/security/x509/X509Certificate.java
```diff
@@ -141,4 +141,26 @@ public String getPEM() {
...
+		String pem = this.pem;
+		int begin = pem.indexOf("-----BEGIN CERTIFICATE-----");
+		if (begin >= 0) {
+			int end = pem.indexOf("-----END CERTIFICATE-----", begin);
+			if (end > begin) {
+				pem = pem.substring(begin, end + "-----END CERTIFICATE-----".length());
+			}
+		}
+		return pem
+				.replace("-----BEGIN CERTIFICATE-----", "")
```
kuntzed: wir haben dafür konstanten, siehe X509Parser.BEGIN_CERTIFICATE und ich finde es unnötig, die erst mit rauszuschneiden, nur um sie dann zu löschen.

kuntzed (approved):

